PR #6, opened 2021-09-08
Alpha

jdheeter:
Merge latest changes to master.
Ignore the github workflows files they are outdated.


.github/workflows/ci-tests.yml
```diff
@@ -23,7 +23,7 @@ jobs:
       if: matrix.platform == 'ubuntu-latest'
       run: |
         sudo apt-get update
-        sudo apt-get install -y --no-install-recommends libwebkit2gtk-4.0-dev
+        sudo apt-get install -y --no-install-recommends libwebkit2gtk-4.0-dev libappindicator3-dev
```
nazar-pc, line 26: And readme too

src/lib/client.ts
```diff
@@ -0,0 +1,218 @@
+import { ApiPromise, WsProvider } from '@polkadot/api';
+import * as event from "@tauri-apps/api/event"
+import { reactive } from 'vue'
+import { LocalStorage } from 'quasar'
+const tauri = { event }
+import { Header } from "@polkadot/types/interfaces/runtime"
```
1devNdogs, line 6: Header is not used

README.md
```diff
@@ -10,9 +10,9 @@ Built with Quasar + Tauri
 
 ### Linux
 
-On Linux you'll have to install development version of `libwebkit2gtk-4.0` package, which can be done on Ubuntu with:
+On Linux you'll have to install development version of `libwebkit2gtk-4.0` and `libappindicator3-dev` package, which can be done on Ubuntu with:
 ```bash
-sudo apt-get install --no-install-recommends libwebkit2gtk-4.0-dev
+sudo apt-get install -y --no-install-recommends libwebkit2gtk-4.0-dev libappindicator3-dev
```
nazar-pc: Don't put `-y` into automated instructions, user needs to confirm the action, this is not a CI.

src/components/plotCard.vue
```diff
@@ -0,0 +1,46 @@
...
+    plot: { type: Object, required: true },
+    config: { type: Object, required: true },
+  },
+})
+</script>
+
+
+
+<style>
+</style>
```
nazar-pc, line 46: Please configure your IDE to always add new line at the end of the file.

tsconfig.json
```diff
@@ -5,6 +5,7 @@
     "baseUrl": ".",
     "noImplicitAny": false,
     "allowJs": false,
+    "strictNullChecks": false,
```
nazar-pc, line 8: Please enforce strict mode everywhere possible. Including this option and `noImplicitAny` or else this codebase is no different from JavaScript and is very easy to break.

nazar-pc (changes requested):

src/components/plotCard.vue
jdheeter: done

tsconfig.json
jdheeter: resolved

src/lib/osUtils/applescript.ts
```diff
@@ -1,19 +1,19 @@
 
 import * as shell from "@tauri-apps/api/shell"
-
+interface ReturnData { stdout: string[], stderr: string[] }
 export async function execString(command: string) {
```
nazar-pc, line 4: This function has no return type, does it really compile? It shouldn't, every standalone function must have type specified explicitly. If it compiler doesn't complain, it should be tweaked to, probably with `noImplicitReturns: true` in tsconfig.json.

src/lib/util.ts
```diff
@@ -10,7 +10,7 @@ import * as native from './native'
 
 import * as bcrypt from 'bcryptjs'
 
-export const random = (min, max) => Math.floor(Math.random() * (max - min)) + min;
+export const random = (min: number, max: number) => Math.floor(Math.random() * (max - min)) + min;
 interface propsType {
   [index: string]: any
```
nazar-pc, line 15: Is it actually completely untyped in Tauri?

src/lib/native.ts
```diff
@@ -23,34 +24,37 @@ export interface AutoLaunchType {
...
+  protected autoLauncher: osAL = nullAL
+  appName: string = 'app'
+  appPath: string = ''
+  enabled: boolean = false
+  async isEnabled(): Promise<boolean> {
+    const result = await this.autoLauncher.isEnabled(this.appName)
+    this.enabled = result
+    return result
+  }
+  async enable(): Promise<any> {
```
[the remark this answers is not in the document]
jdheeter: I don't currently need any return value but I may implement a return value in the future so I set it to any as a placeholder. I can remove it if you like.

src/lib/global.ts
```diff
@@ -2,12 +2,12 @@ import { reactive } from "vue";
...
 import getLang from "../loc/lang"
 import { Client, ClientType } from "src/lib/client"
-import { autoLaunch, AutoLaunchType } from "src/lib/native"
+import { AutoLauncher } from "src/lib/native"
 
 
 let text: any
 let client!: ClientType
-let launchOnBoot!: AutoLaunchType
+let launchOnBoot!: AutoLauncher
```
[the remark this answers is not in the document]
jdheeter: They are async initialized in the global state object in the mutations. I could convert it to a class with async initialize function to be more readable.

src/lib/util.ts
jdheeter: can you explain?

src/lib/native.ts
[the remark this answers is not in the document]
nazar-pc: `any` should almost never be used. This is because it can be used in place of ANY type, breaking the whole idea of typing in TypeScript, making it useless since you break invariant in one place and the whole app can't be trusted anymore.

If you don't have a value to return, make it `Promise<void>` to express that explicitly. If you change your mind in the future you will have to update the signature and code that relied on it to be `Promise<void>` will stop compiling, which is what you want from a typed language.

src/lib/osUtils/applescript.ts
jdheeter: Sure I can clean it up more.

src/lib/global.ts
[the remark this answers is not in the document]
nazar-pc: I have never seen this approach before. It looks confusing and seems like a dead code to me. If you can make it more clear I'll be thankful.

src/lib/client.ts
```diff
@@ -81,14 +81,14 @@ function getStoredBlocks(): FarmedBlock[] {
     for (let [num, block] of Object.entries(blocks)) {
       mined.push(block as FarmedBlock)
     }
-  } catch (error) {
+  } catch (error: any) {
```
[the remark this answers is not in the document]
jdheeter: throws an error if you try to access error like error.toString() since error is 'unknown' type by default.

src/lib/util.ts
nazar-pc: This says that an object is expected that can hold anything as a value of a key. Then this interface is used in `Dialog.create()` call below. I have a strong suspicion it doesn't actually accept anything, but rather have a specific constraint on values (like strings) or maybe even has a specific requirements to the keys supported. In many cases you can "extract" the type of that parameter from the library too instead of resorting to `any` that erases all of the type information.

src/lib/client.ts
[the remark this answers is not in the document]
nazar-pc: Which in turn means that by slapping `any` you create a time bomb that will blow up once the library throws something that doesn't have `.toString()` method on it. Your need to either specify a concrete type (if there are guarantees that nothing else can possibly be thrown out of there) or write code in such a way that can handle any value.

In this particular case just log the variable itself, `.toString()` will be called on it automatically if implemented, otherwise platform-specific behavior will get at least something printed to the console without causing runtime error.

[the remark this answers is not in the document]
jdheeter: Sure I can do that

src/lib/global.ts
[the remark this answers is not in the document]
jdheeter: will do

isSerge:
could you, please, create a separate PR with eslint and format the code? can just extend `eslint:recommended` for now

src/lib/util.ts
jdheeter: cleaned up

src/lib/osUtils/applescript.ts
```diff
@@ -0,0 +1,30 @@
...
+import * as shell from "@tauri-apps/api/shell"
+import { ChildReturnData } from "../types";
+
+export async function execString(command: string): Promise<ChildReturnData> {
+  let args = ['-e', command]
+  console.log("Executing applescript:", args);
+
+  const child = new shell.Command('osascript', args)
+  return new Promise((res) => {
+    let returnData = <ChildReturnData>{ stdout: [], stderr: [] }
```
nazar-pc, line 11: If you need to write something like that, I think following is more idiomatic:
```suggestion
    const returnData: ChildReturnData = { stdout: [], stderr: [] }
```

src/lib/util.ts
```diff
@@ -74,7 +75,8 @@ export const config = {
   },
   async update(newData: ConfigFile, dir?: string) {
     let config = await this.read(dir).catch(console.error) || emptyConfig
-    for (const [key, value] of Object.entries(newData)) {
+    for (let [key, value] of Object.entries(newData)) {
+      key = key as string
       config[key] = Object.assign(config[key], value)
```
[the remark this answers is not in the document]
nazar-pc: This wasn't actually resolved, un-resolving

src/lib/native.ts
```diff
@@ -0,0 +1,90 @@
...
+    console.log('get_this_binary', this.appPath);
+    console.log(this.appPath);
+    if (os == 'Darwin') this.autoLauncher = macAL
+    else if (os == 'Windows_NT') this.autoLauncher = winAL
+    else this.autoLauncher = linAL
+    this.enabled = (await this.isEnabled())
+  }
+}
+
+export async function execString(executable: string, args: string[] | string) {
```
nazar-pc, line 57: This function returns `boolean` and while it isn't specified, nothing complains. This should fail at compile time complaining about wrong types, the correct return type of this function is `Promise<boolean>`. The same comment applies to other functions in this file. Without such checks, we essentially don't have TypeScript and can return random things from random functions with wrong signatures.

Let me know if this is already taken care of by eslint in the other PR (I was using tslint for similar purpose before, but it is discontinued, I'm sure there are similar options for eslint).

src/lib/global.ts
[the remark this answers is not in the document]
nazar-pc: This wasn't resolved either despite marked as such, un-resolving

src/lib/util.ts
```diff
@@ -73,7 +74,7 @@ export const config = {
   },
   async update(newData: ConfigFile, dir?: string) {
     let config = await this.read(dir).catch(console.error) || emptyConfig
-    for (const [key, value] of Object.entries(newData)) {
+    for (let [key, value] of Object.entries(newData)) {
```
nazar-pc, line 77: Should probably be `const` again, ideally this should also be linted

src/pages/SetupPlot.vue
```diff
@@ -109,71 +105,65 @@ export default defineComponent({
...
-    },
-    safeAvailableGB(): number {
-      return this.driveStats.freeBytes / 1e9
-    },
-    utilizedGB(): number {
-      return parseFloat((this.totalDiskSizeGB - this.safeAvailableGB).toFixed(2))
-    },
-    freeGB(): number {
-      return parseFloat((this.safeAvailableGB - this.allocatedGB).toFixed(2))
+    stats(): any {
```
nazar-pc, line 124: This and other functions in this file should return something meaningful instead of `any`.

This isn't serious if you have a very specific data structure here and writing `any` just to make compiler shut up.

src/pages/SetupPlot.vue
```diff
@@ -109,71 +105,65 @@ export default defineComponent({
...
         if (this.plotDirectory == this.defaultPath) return (this.validPath = true)
-        this.validPath = await util.native.dirExists(val)
+        this.validPath = await native.dirExists(val)
         if (this.validPath) await this.updateDriveStats()
+        return
       }, 500)
     )
   },
   computed: {
     chartData(): any {
```
nazar-pc, line 115: This not only need a proper type, but unless it is an object with descriptive keys, it also needs a comment explaining what is being returned from here so that it is readable from function signature + its comment.

nazar-pc (commented):
Thank you for improved typing, but I still have concerns and a few nitpicks. I didn't really check business logic to be honest.

nazar-pc:
Cleaned up Justin's commits

src/lib/global.ts
```diff
@@ -5,29 +5,49 @@ import { Client, ClientType } from "src/lib/client"
...
+  changeLang(newLang: string): void
+  loadLangData(): void
+}
+export class Global {
+  data = reactive({ status: { state: "loading", message: "loading" }, loc: { selected: 'en', text, }, client, launchOnBoot })
+  mutations: GlobalMutations = {
+    initClient: this.initClient,
+    initLaunchOnBoot: this.initLaunchOnBoot,
+    changeLang: this.changeLang,
+    loadLangData: this.loadLangData
```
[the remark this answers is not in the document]
jdheeter: Here I am just organizing the protected functions in the class into an public object. So inside the class `this.initClient` is a reference to `initClient()` on line 27.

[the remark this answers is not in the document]
nazar-pc: `initClient()` doesn't touch `this.mutations` unless I'm missing something.

src/lib/global.ts
```diff
@@ -5,29 +5,49 @@ import { Client, ClientType } from "src/lib/client"
...
+  }
+  protected async initClient() {
+    this.data.client = null
     const client = await Client()
```
nazar-pc, line 29: This is somewhat confusing too. I'd expect either `new Client()` or `await Client.init()`, but not `await Client()`

src/lib/native.ts
```diff
@@ -34,11 +33,11 @@ export class AutoLauncher {
     this.enabled = result
     return result
   }
-  async enable(): Promise<any> {
+  async enable(): Promise<void | ChildReturnData> {
```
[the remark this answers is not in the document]
jdheeter: Honestly I'm not using the stdout/stderr right now in the code when calling these functions but I figured in the future I may want to throw an error or status update based on that data. Sometimes there can be a failure ( thought I haven't seen that in my testing) in which case it would be good to be able to detect that and alert the user. This code was improved already in the #13 branch so I won't modify it here.

src/lib/osUtils/applescript.ts
jdheeter: Sure, I will update that. Is there any advantage to typing the variable one way or the other?

src/lib/global.ts
jdheeter: It's because Client is not a class. I can convert it to a class but I don't want to modify it until after we have merged in the changes from Leo.

src/lib/osUtils/applescript.ts
nazar-pc: I honestly didn't test in TypeScript, but generally `const v = <T>{}` is casting and `const v: T = {}` is not, so in second case it will definitely complain if object isn't of the correct shape. I guess your option should also work, but in practice you don't even need it in majority of cases if the function signature is typed, TypeScript will check whether it matches expected interface and should:
1) Infer the type automatically
2) Complain if the type is incorrect

The most often case when you need an explicit type is when you have type `'a' | 'b'` and TypeScript infers wider type `const v: string = 'a'`, but this isn't the case here.

src/lib/global.ts
nazar-pc: I agree, let's have it as is for now

src/pages/SetupPlot.vue
jdheeter: Sure I can type it

src/pages/SetupPlot.vue
jdheeter: anon function inside debouce function so returns null, only added a return because linting error

nazar-pc: Hold on resolving this, I don't quite understand what you said. This function returns an array with 3 elements, what `null` are you talking about?

package.json
```diff
@@ -36,6 +36,8 @@
     "@types/ms": "^0.7.31",
     "@types/node": "^16.3.3",
     "@types/vue": "^2.0.0",
+    "i": "^0.3.6",
+    "npm": "^7.24.0",
```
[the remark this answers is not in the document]
jdheeter: removed

src/pages/SetupPlot.vue
jdheeter: Sorry I'm looking at the function above the one you are talking about. I have resolved the issue in the latest commit.

src/lib/osUtils/applescript.ts
jdheeter: ok good to know

src/lib/native.ts
jdheeter: I agree that many functions in this file could be cleaned up. There is no linting or compile errors so I'm not sure what setting we need to have to formally mark these as errors.  I think most issues resolved in eslint branch #9

src/lib/global.ts
```diff
@@ -17,7 +21,8 @@ export interface GlobalMutations {
   loadLangData(): void
 }
 export class Global {
-  data = reactive({ status: { state: "loading", message: "loading" }, loc: { selected: 'en', text, }, client, launchOnBoot })
+  private client!: ClientType
+  data = reactive({ status: { state: "loading", message: "loading" }, loc: { selected: 'en', text }, launchOnBoot: new AutoLauncher, client: this.client })
```
nazar-pc, line 25: I do not understand what this does either.

You just moved `client` variable to private property that is incorrectly typed as `ClientType` (there is no private constructor to enforce this invariant and no methods in the class that assign anything to the field in general), in fact it is guaranteed to always be `undefined` and nothing else.

nazar-pc: Hm, I think I do understand. You want to have async initialization, but you also want to pretend it is initialized synchronously for convenient use elsewhere. But it will result in strange bugs because the invariant you rely on are can't be guaranteed and you have to call `init()` for it to actually be true. Am I right?

nazar-pc: Hm... it must already be broken, `this.initClient()` in `init()` is commented-out, so there is no code in this file that can guarantee `data.client` is actually anything other than `undefined`. Does it actually work?

jdheeter: I can't initialize it in the constructor because it's async. I have to have some reference to it inside data for other code to be valid. `this.initClient()` is commented out for debug purposes. I appreciate the scrutiny but in this case we are nitpicking code that isn't meant to be close to finalized. Once we merge the latest branches together I will refactor Client into a class in which case it can be initialized using `new Client` similar to how I handle `launchOnBoot`.

nazar-pc: If client isn't used, let's remove it from `data`. Otherwise as far as I understand it this is guaranteed to be broken, I don't see a working code path.

As to async constructors, you can have them:
```typescript
class T {
  private construct(public readonly client) {}
  public static async init(): Promise<T> {
    const client = await Client.init();
    return new T(client);
  }
}
```

This way you physically can't construct invalid invariant of `T` because its constructor is private and the only way to ever get its instance is to go through `T.init()`.

src/lib/client.ts
```diff
@@ -105,117 +104,75 @@ function clearStored() {
...
-                const [type, data] = log.asPreRuntime
-                if (type.toString() === 'POC_') {
-                  const poCPreDigest: PoCPreDigest = api.registry.createType('PoCPreDigest', data)
-                  const solution: Solution = api.registry.createType('Solution', poCPreDigest.solution)
-                  const farmerId: FarmerId = api.registry.createType('FarmerId', solution.public_key)
-                  console.log('farmerId: ', farmerId.toString())
-                }
+export class Client {
+  wsProvider = new WsProvider
+  api: ApiPromise = new ApiPromise
```
nazar-pc, line 109: I believe this will still not properly work until `Client.init()` is called. Please put a big TODO to refactor this is a way that doesn't have this issue.

And another TODO on top of this class to add visibility to all methods/properties, most of this stuff shouldn't be public.

src/lib/global.ts
```diff
@@ -1,46 +1,28 @@
...
-  protected async initClient() {
-    const client = await Client()
-    this.data.client = client
-  }
-  protected async initLaunchOnBoot() {
-    this.data.launchOnBoot.appName = 'app'
-    await this.data.launchOnBoot.init()
-  }
-  protected async changeLang(newLang: string) {
+  client = new Client
```
nazar-pc, line 8: Add big TODO to refactor this to either protect invariants (it exposes `.client` that is in a broken state until you call `.client.init()` and `await` on it or better yet refactor completely and use dependency injection, such that users of this don't even run before client is ready to accept requests.

src/pages/Dashboard.vue
```diff
@@ -51,9 +51,10 @@ export default defineComponent({
     }
     let unsubscribe: VoidFn = () => {}
     let clientData: ClientData = emptyData
-    return { lang, config, network, plot, global: global.data, globalState, expanded: false, util, loading: true, unsubscribe, clientData }
+    return { lang, config, network, plot, global: global.data, client: global.client, globalState, expanded: false, util, loading: true, unsubscribe, clientData }
   },
   async mounted() {
+    this.client.init()
```
nazar-pc, line 57: Add TODO to remove this once invariants are protected. Someone WILL use `this.client` at some point when they forget to call `this.client.init()` in another component and things WILL break in a racy way, I have seen enough applications that are broken like this.

src/pages/Dashboard.vue
```diff
@@ -75,16 +76,16 @@ export default defineComponent({
...
         return val.blockReward + val.feeReward + agg
       }, 0)
     },
   },
   unmounted() {
     this.unsubscribe()
-    this.global.client?.do?.blockSubscription.stop()
-    this.global.client?.data?.farming.events.off("farmedBlock", this.farmBlock)
+    this.client.do?.blockSubscription.stop()
+    this.client.data?.farming.events.off("farmedBlock", this.farmBlock)
```
nazar-pc, line 88: Both `this.client.do` and `this.client.data` always exist, remove `?` to prevent confusion.

src/pages/Dashboard.vue
```diff
@@ -102,8 +103,8 @@ export default defineComponent({
   },
   methods: {
     async testClient() {
-      this.global.client?.do?.runTest()
-      this.global.client?.data?.farming.events.on("farmedBlock", this.farmBlock)
+      this.client?.do?.blockSubscription.runTest()
+      this.client?.data?.farming.events.on("farmedBlock", this.farmBlock)
```
nazar-pc, line 107: Same here, all `?` should be removed, none of the fields are optional as far as I see.

src/lib/client.ts
```diff
@@ -104,24 +104,18 @@ function clearStored() {
...
-  data = emptyData
-  clearTauriDestroy: event.UnlistenFn = () => { }
-  unsubscribe: event.UnlistenFn = () => { }
-  constructor() {
-    this.clientData.farming.farmed = this.farmed
-  }
-  async init() {
-    this.api = await ApiPromise.create({ provider: this.wsProvider, types: customTypes })
-  }
+  protected wsProvider = new WsProvider
```
nazar-pc, line 109: Just FYI, `protected` should only be used when you intend the property or method to ever be accessible to the class that extends it. Otherwise, it should be `private`.

nazar-pc (approved):
Let's move forward, but this still needs more work IMO

1devNdogs (approved):

PR #10, opened 2021-09-21
Entitlements & v0.0.3 Publish Update

ImmaZoni:
Entitlements are established, updated publish command to version v0.0.3, and set as release to build to proper code signing, notarization, and entitlements.


nazar-pc (approved):
Is it expected that 6 commits of changes result in just 1 line difference? Maybe you can rebase the PR to avoid a bunch of useless commits in the history?

PR #11, opened 2021-09-21
Alpha custom types

1devNdogs:
This PR add custom types for requested features: Peers and Block Author information, so they can be used by John in the farmer interface.
s.

- I have manually imported custom types from a JSON file, generated by subspace.js project.
- Created types interfaces to inject FarmerId and PocPreDigest to the Polkadot API promise.
- Added/Replaced Block author information in the console.log. So they can be used now by John needs.


src/lib/client.ts
```diff
@@ -135,42 +136,43 @@ export const Client = async () => {
           client.do.blockSubscription.stop()
         },
         async start() {
-          if (!client.api) throw (Error("Api Missing, can't start block subscription yet."))
+          if (!client.api) throw (Error(`Api Missing, can't start block subscription yet.`))
```
nazar-pc, line 139: Why change to template string literal if there is no interpolation?

nazar-pc (approved):
Last commit shouldn't have been here. It is better not to mix formatting PR with feature PR like this one, it makes PR bigger than necessary and harder to review for no reason.

Also please hold on merging it, we need to merge `alpha` to `main` first to avoid re-reviewing these feature changes and formatting updates alongside with changes in `alpha` many more times.

src/lib/client.ts
1devNdogs: Should i just use back slash on the **can't** single quote?

nazar-pc: Hm, is it required by linter? If not, I'd leave the way it was, but it is a minor nitpick.

jdheeter (commented):
looks fine: merged

nazar-pc:
I explicitly asked not to merge this before `alpha` is merged into `main` either. You just made `alpha` over 100 lines bigger again :disappointed: 

The whole idea was to have reasonably sized PRs that can be merged one by one instead of having a huge one that is hard to review, hard to resolve requested changes and hard to merge. By merging more stuff into `alpha` we make it even bigger than it was.

Every time a piece of code is changed in existing PR, reviewer needs to:
1) Analyze changes in isolation
2) Analyze the whole PR with those changes on top

^ is counter-productive for everyone involved.

jdheeter:
Sorry for the confusion, In order to resolve your comments on the alpha branch I need the changes from this branch to be merged. Otherwise there will be massive conflicts merging this branch later.

PR #13, opened 2021-09-22
Windows autolaunch

jdheeter:
works great on windows but needs work to compile on other platforms


nazar-pc:
Please re-request review once it works cross-platform

PR #16, opened 2021-09-29
Windows autolaunch

jdheeter:
Cross platform compilation is handled correctly now.

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -38,6 +38,7 @@ dependencies = [
  "serde_json",
  "tauri",
  "tauri-build",
+ "winreg",
 ]
 
 [[package]]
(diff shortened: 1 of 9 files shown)


nazar-pc:
Can you squash commits, please? There are multiple that are a part of the same exact thing, it would make history much smaller and cleaner.

jdheeter:
recent commits squashed

nazar-pc:
> recent commits squashed

Well, when you squash commits, the number of commits is supposed to decrease. Here I would expect 2 or maybe 3 commits total. But it only got bigger :disappointed:

jdheeter:
Usually I squash during merge of the PR instead of removing old commits but yes I can make it happen.

nazar-pc:
You have a few distinct changes here, so it makes sense to me to have 2 or 3 commits with independent changes. Instead of squashing everything into one during merge.

PR #18, opened 2021-09-29
Windows autolaunch

jdheeter:
Created fresh PR with clean history after squash

### DIFF
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -16,7 +16,8 @@ tauri-build = { version = "1.0.0-beta.2" }
 serde_json = "1.0"
 serde = { version = "1.0", features = [ "derive" ] }
 fs2 = "0.4.3"
-# winreg = "0.10.1"
+[target.'cfg(windows)'.dependencies]
+winreg = "0.10.1"
 
 [dependencies.tauri]
 version = "1.0.0-beta.8"
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -16,6 +16,7 @@ tauri-build = { version = "1.0.0-beta.2" }
 serde_json = "1.0"
 serde = { version = "1.0", features = [ "derive" ] }
 fs2 = "0.4.3"
+
 [target.'cfg(windows)'.dependencies]
 winreg = "0.10.1"
 
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -2,27 +2,29 @@
   all(not(debug_assertions), target_os = "windows"),
   windows_subsystem = "windows"
 )]
-use std::path::PathBuf;
+
+#[cfg(target_os = "windows")]
+mod windows;
 
 use serde::Serialize;
-
+use std::path::PathBuf;
 use tauri::{
   api, CustomMenuItem, Manager, SystemTray, SystemTrayEvent, SystemTrayMenu, SystemTrayMenuItem,
   WindowEvent,
 };
 
 #[derive(Serialize)]
-struct S {
+struct DiskStats {
   free_bytes: u64,
   total_bytes: u64,
 }
 
 #[tauri::command]
-fn get_disk_stats(dir: String) -> S {
+fn get_disk_stats(dir: String) -> DiskStats {
   println!("{}", dir.to_string());
   let free: u64 = fs2::available_space(&dir).expect("error");
   let total: u64 = fs2::total_space(&dir).expect("error");
-  return S {
+  return DiskStats {
     free_bytes: free,
     total_bytes: total,
   };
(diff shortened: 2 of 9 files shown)


src-tauri/Cargo.toml
```diff
@@ -17,6 +17,8 @@ serde_json = "1.0"
 serde = { version = "1.0", features = [ "derive" ] }
 fs2 = "0.4.3"
 # winreg = "0.10.1"
```
nazar-pc, line 19: Let's remove this comment then

src-tauri/src/main.rs
```diff
@@ -5,11 +5,18 @@
...
 
 use serde::Serialize;
+
 use tauri::{
   api, CustomMenuItem, Manager, SystemTray, SystemTrayEvent, SystemTrayMenu, SystemTrayMenuItem,
   WindowEvent,
 };
 
+#[cfg(target_os = "windows")]
+extern crate winreg;
```
nazar-pc, line 15: Remove this, it is unnecessary in modern Rust.

src-tauri/src/main.rs
```diff
@@ -5,11 +5,18 @@
...
+
+#[cfg(target_os = "windows")]
+use {winreg::enums::*, winreg::RegKey};
```
nazar-pc, line 18:
```suggestion
#[cfg(target_os = "windows")]
use winreg::{enums::*, RegKey};
```

src-tauri/src/main.rs
```diff
@@ -32,6 +39,69 @@ fn get_this_binary() -> PathBuf {
...
+  let mut result: String = "success".to_string();
+  match val {
+    Ok(()) => {}
+    Err(error) => {
+      println!("There was an error: {}", error);
+      result = error.to_string();
+    }
+  }
+  return result;
+}
```
nazar-pc, line 104: I would move these 3 into separate module and put one `#[cfg(target_os = "windows")]` on the whole module. That would be cleaner and easier to follow.

src-tauri/src/main.rs
```diff
@@ -32,6 +39,69 @@ fn get_this_binary() -> PathBuf {
...
+#[cfg(target_os = "windows")]
+#[tauri::command]
+fn winreg_set(sub_key: String, set_key: String, value: String) -> String {
+  let hkcu = RegKey::predef(HKEY_CURRENT_USER);
+  // let subkey = hkcu.open_subkey(sub_key).unwrap();
+  let (subkey, disp) = hkcu.create_subkey(&sub_key).unwrap();
+  match disp {
+    REG_CREATED_NEW_KEY => println!("A new key has been created"),
+    REG_OPENED_EXISTING_KEY => println!("An existing key has been opened"),
+  }
```
nazar-pc, line 72: Are these just debug messages?

src-tauri/src/main.rs
```diff
@@ -32,6 +39,69 @@ fn get_this_binary() -> PathBuf {
...
+  let subkey = hkcu.open_subkey(sub_key).unwrap();
+  let val = subkey.get_value(value);
+  let result: String;
+  match val {
+    Ok(content) => {
+      println!("File content: {}", content);
+      result = content;
+    }
+    Err(error) => {
+      println!("There was an error: {}", error);
```
nazar-pc, line 55: For stderr:
```suggestion
      eprintln!("There was an error: {}", error);
```

src/lib/autoLaunch/winAutoLaunch.ts
```diff
@@ -1,56 +1,29 @@
...
-    //   return regKey.set(appName, Winreg.REG_SZ, `\"${pathToAutoLaunchedApp}\"${args}`, function (err) {
-    //     if (err != null) { return reject(err); }
-    //     return resolve();
-    //   });
-    // });
+const winAutoLaunch = {
+  // eslint-disable-next-line no-unused-vars
+  async enable({ appName, appPath, hidden }: AutoLaunchParams): Promise<ChildReturnData> {
+    let returnVal = <ChildReturnData>{ stdout: [], stderr: [] }
+    const result = <string>(await native.winregSet(subKey, appName, appPath))
```
nazar-pc, line 10: `native.winregSet` already has return type specified, no need to repeat it, much easier to read this way
```suggestion
    const result = await native.winregSet(subKey, appName, appPath)
```

src/lib/native.ts
```diff
@@ -86,3 +88,26 @@ export async function driveStats(dir: string): Promise<DriveStats> {
...
+  const result = (await tauri.invoke('winreg_set', { subKey, setKey, value })) as string
+  console.log(result);
+  return result
+}
+
+export async function winregDelete(subKey: string, setKey: string): Promise<string> {
+  const result = (await tauri.invoke('winreg_delete', { subKey, setKey })) as string
+  console.log(result);
+  return result
+}
```
nazar-pc, line 108: These should really be called differently all the way through. The fact that implementation detail (library `winreg`) leaks all the way here is not pretty.

Also is `console.log` really needed here?

nazar-pc (commented):
History is better, but third commit is out of place.

A few minor comments. None of them are critical, but it is not tidy the way it is right now.

src/lib/native.ts
jdheeter: How would you suggest to structure these functions? These are mostly just a wrappers to add type checking around the tauri.invoke functions that call my rust functions.

nazar-pc: Both these functions and those that are in Rust shouldn't have `winreg` prefix.

For instance, instead of `function winregGet(subKey: string, value: string): Promise<string>` I would expect `function isWindowsAutostartEnabled(): Promise<boolean>` (returns boolean if everything is fine and throws `Error` if something went wrong).

nazar-pc:
There is 11 commits again, what is going on?

src/lib/native.ts
jdheeter: The Rust functions and ts wrappers are meant to be general windows registry utility functions, they can be used to read/write any registry data, which will be useful for more windows functionality in the future. I can refactor into special case functions but imo there is no benefit to doing so.

nazar-pc: Having direct access from frontend into Windows Registry or being able to run arbitrary executables is a security issue waiting to happen, also it makes it nearly impossible to audit what frontend can or can't do. Clear purpose-built and restricted methods are easier to use, easier to audit and easier to keep secure.

nazar-pc:
I'm tired :disappointed: 
I will clean it up later myself.

jdheeter:
Old version of branch was on another machine, everything synced up now.

src/lib/native.ts
jdheeter: Your comment also applies to the applescript utility functions so I made a TODO to refactor winreg and applescript related functions to make them more use case specific. I also have functions that can read/write files to arbitrary directories that could be refactored.

nazar-pc:
@jdheeter can you check if it still works as expected?

nazar-pc (approved):

jdheeter:
ran locally and works

PR #19, opened 2021-09-30
Alpha block author and peers information

1devNdogs:
Contain changes to show block author and peers information.

Added block author from polkadot/api, as FarmerId

Added peers information to network card data.


jdheeter (approved):
Looks good. I agree the client initialization could be cleaner. We could implement dependency injection but probably outside the scope of this PR. The TODOs seems sufficient to highlight potential issues.

src/lib/client.ts
```diff
@@ -106,16 +106,19 @@ function clearStored() {
...
+  data = reactive(emptyData);
+
   status = {
     farming: () => { }, // TODO return some farming status info
     plot: () => { }, // TODO return some plot status info
-    net: () => { } // TODO return some net status info
+    net: () => {
+      return this.api.rpc.system.peers()
+    } // TODO return some net status info
   }
```
nazar-pc, line 122: Can we have some type information here? I don't see anything and wonder how it even compiles :thinking:

jdheeter: The return type is inferred by ts similar to assigning a variable without explicitly defining the variable type. I looked for an option for force explicit return type that could be placed in tsconfig but didn't see it. I do see an option to type check function returns in eslint so I'll try and enable that.
image
https://github.com/bradzacher/eslint-plugin-typescript/blob/master/docs/rules/explicit-function-return-type.md

nazar-pc: There is a big difference between variables and functions/methods. For variables it is fine to infer types, but for functions input and output types should be defined explicitly (the exception is anonymous callbacks whose type is defined in the argument already).

jdheeter: I got the eslint change implemented and I'm pushing the changes on another branch right now.

jdheeter: Ok I implemented improved checking in another branch that will force functions to have an explicit return type defined.
#20

jdheeter (approved):
ready to merge

nazar-pc (approved):

PR #20, opened 2021-10-03
Eslint update

jdheeter:
I improved the eslint config to catch more issues. Includes `@typescript-eslint/explicit-function-return-type` rule which will throw an error if a function return type is not explicitly defined. There is a bug where this specific rule does not work in .vue files. I don't think this is an issue since most functions in vue files are `computed` which requires explicit return anyways or `methods` which almost always return void. 

They said the bug was fixed I'm not seeing it, not a big issue to me anyways. 
https://github.com/typescript-eslint/typescript-eslint/issues/524


nazar-pc (approved):
Looks fine, but linting isn't checked in CI unless I'm missing something, so it wouldn't catch any inconsistencies automatically

1devNdogs (dismissed):

jdheeter:
I added lint to the build action so if lint fails the build will fail now.

nazar-pc (approved):

PR #21, opened 2021-10-03
Fix dependencies installation, rename action

nazar-pc:


ImmaZoni (approved):
Looks good, dependency updates worked and allowed a successful build on my version of Ubuntu.

Issue #3, opened 2021-08-14
thread 'main' panicked at 'data.len()'


ImmaZoni:
## Describe The Issue
After installing the .deb file i am unable to run on Ubuntu, receiving the following error:
`thread 'main' panicked at 'data.len() must fit the width, height, and row_stride', /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/gdk-pixbuf-0.9.0/src/pixbuf.rs:57:13`

Showing the Full Rust Backtrace gives us the following details. 

```RUST_BACKTRACE=full subspace-farmer-demo
thread 'main' panicked at 'data.len() must fit the width, height, and row_stride', /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/gdk-pixbuf-0.9.0/src/pixbuf.rs:57:13
stack backtrace:
   0:     0x55a873d519a0 - std::backtrace_rs::backtrace::libunwind::trace::h34055254b57d8e79
                               at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/std/src/../../backtrace/src/backtrace/libunwind.rs:90:5
   1:     0x55a873d519a0 - std::backtrace_rs::backtrace::trace_unsynchronized::h8f1e3fbd9afff6ec
                               at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x55a873d519a0 - std::sys_common::backtrace::_print_fmt::h3a99a796b770c360
                               at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/std/src/sys_common/backtrace.rs:67:5
   3:     0x55a873d519a0 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h32d1f94a80615d18
                               at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/std/src/sys_common/backtrace.rs:46:22
   4:     0x55a873d77c7c - core::fmt::write::h306731c068f7162c
                               at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/core/src/fmt/mod.rs:1110:17
   5:     0x55a873d4ae65 - std::io::Write::write_fmt::hd2fa90334eee2a21
                               at /rustc/a178d0322ce20e33eac124758e837cbd80a6f633/library/std/src/io/mod.rs:1588:15
[...]

PR #23, opened 2021-10-04
implement linux autostart

jdheeter:
I'm not able to test locally yet but will try tomorrow.


src/lib/autoLaunch/linuxAutoLaunch.ts
```diff
@@ -1,29 +1,44 @@
+import { AutoLaunchParams, ChildReturnData } from '../types'
+import * as native from '../native'
+import * as fs from "@tauri-apps/api/fs"
 
-module.exports = {
+const linuxAutoLaunch = {
```
nazar-pc, line 5: Can we have this as a class that implements an interface shared for all 3 platforms?

src/lib/autoLaunch/linuxAutoLaunch.ts
```diff
@@ -1,29 +1,44 @@
...
+    try {
+      await fs.readTextFile(this.getFilePath(appName))
+      return true
+    } catch (error) {
+      console.error(error)
+      return false
+    }
+  },
+  getDirectory(): string { return '~/.config/autostart/'; },
+  getFilePath(appName: string): string { return `${this.getDirectory()}${appName}.desktop`; }
```
nazar-pc, line 42: I don't think these are supposed to be public API.

src/lib/autoLaunch/linuxAutoLaunch.ts
```diff
@@ -1,29 +1,44 @@
...
-};
+    const contents = `
+    [Desktop Entry]
+    Type=Application
+    Version=1.0
+    Name=${appName}
+    Comment=${appName}startup script
+    Exec=${appPath}${hiddenArg}
+    StartupNotify=false
+    Terminal=false
```
nazar-pc, line 19:
```suggestion
[Desktop Entry]
Type=Application
Name=${appName}
Comment=${appName} startup script
Exec=${appPath}${hiddenArg}
Icon=PATH_TO_APP_ICON
```

src/lib/autoLaunch/linuxAutoLaunch.ts
jdheeter: The class that implements the cross platform functions is in native.ts but I agree that things could be organized in a better way.

src-tauri/src/main.rs
```diff
@@ -100,6 +100,8 @@ fn main() {
     windows::winreg_get,
     windows::winreg_set,
     windows::winreg_delete,
+    get_this_binary,
```
jdheeter, line 103: @nazar-pc I found that this command was not compiled on windows unless I add it here.

src/lib/autoLaunch/linuxAutoLaunch.ts
jdheeter: consolidated in latest commit

src-tauri/src/main.rs
nazar-pc: Yeah, that is correct after I read the implementation in Tauri, thanks!

src/lib/autoLauncher.ts
```diff
@@ -0,0 +1,141 @@
...
+  }
+  async enable(): Promise<void | ChildReturnData> {
+    const child = await this.autoLauncher.enable({ appName: this.appName, appPath: this.appPath, hidden: false })
+    return child
+  }
+  async disable(): Promise<void | ChildReturnData> {
+    const child = this.autoLauncher.disable(this.appName)
+    return child
+  }
+  async init(): Promise<void> {
```
nazar-pc, line 130: The same flawed approach with initialization here, please either make sure it is impossible to instantiate uninitialized `AutoLauncher` or add TODO to fix that.

src/lib/autoLauncher.ts
```diff
@@ -0,0 +1,141 @@
...
+    const exists = loginList.includes(appName)
+    console.log('login Item Exists:', exists);
+    return exists
+  },
+
+}
+
+const winAL = {
+  subKey: "SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\Run",
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
nazar-pc, line 88: Shouldn't you have TODO to support `hidden`? Also if I understand it, better name would be `minimized`.

src/lib/autoLauncher.ts
```diff
@@ -0,0 +1,141 @@
...
+
+    const contents = `
+    [Desktop Entry]
+    Type=Application
+    Version=1.0
+    Name=${appName}
+    Comment=${appName}startup script
+    Exec=${appPath}${hiddenArg}
+    StartupNotify=false
+    Terminal=false
```
nazar-pc, line 40:
```suggestion
[Desktop Entry]
Type=Application
Name=${appName}
Comment=${appName} startup script
Exec=${appPath}${hiddenArg}
Icon=PATH_TO_APP_ICON
```

src/lib/autoLauncher.ts
```diff
@@ -0,0 +1,141 @@
...
+    if (result.search('The system cannot find the file specified.') > -1) return false
+    else return true
+  }
+}
+
+
+export class AutoLauncher {
+  protected autoLauncher: osAL = nullAL
+  appName = 'app'
+  appPath = ''
```
nazar-pc, line 115: Should this field be public? If so why?

src/lib/autoLauncher.ts
jdheeter: Minimized implies that there is a window which is minimized to the taskbar but what this actually means is don't spawn a widow at all. In hidden mode the app would just start in the tray on the respective platform with no window would be spawned until you open the tray menu. Added a TODO to handle this case on windows.

nazar-pc: What describe is that the window is effectively there, it is just not visible until you open the tray menu. I believe it is called `minimized` in most other apps, not `hidden`. In Linux at least there is actually a separate `hidden` attribute in the autostart file which will make it invisible in GUI that manages autostarting apps.

jdheeter: Ok, I'm not sure of the specifics, this nomenclature came from the npm library I refactored which called it  isHiddenOnLaunch so I shortened to hidden. I can change to minimized.
https://github.com/Teamwork/node-auto-launch/blob/master/src/AutoLaunchWindows.coffee

src/lib/autoLauncher.ts
nazar-pc: The changes implemented are not the changes I suggested. Please don't resolve things unless they are absolutely 100% resolved, it is annoying to go through resolved threads and unresolve them over and over again.

jdheeter:
I was not able to test this working in Ubuntu because the app crashes on start #3 . Based on looking at the npm library from electron, it should work. I think safe to merge and check later when Linux build is not crashing.

src/lib/autoLauncher.ts
jdheeter: ah my mistake

nazar-pc: It is more about contents than spacing actually

jdheeter: ok updated

jdheeter:
@nazar-pc I think ready to merge

nazar-pc (approved):

PR #26, opened 2022-01-08
Windows Codesign Support

ImmaZoni:
This PR Adds windows code signing support when building the application. 

Code signing was tested locally and working.


nazar-pc (commented):
Looks fine, but CI needs to pass

nazar-pc (approved):

nazar-pc:
Please squash-merge once CI is green

.github/workflows/ci-tests.yml
```diff
@@ -37,7 +37,7 @@ jobs:
         Set-Content -Path certificate/tempCert.txt -Value $env:WINDOWS_PFX
         certutil -decode certificate/tempCert.txt certificate/certificate.pfx
         Remove-Item –path certificate -include tempCert.txt
-        Import-PfxCertificate -FilePath certificate/certificate.pfx -CertStoreLocation Cert:\LocalMachine\My -Password (ConvertTo-SecureString -String $env:WINDOWS_PFX_PASSWORD -Force -AsPlainText)
+        Import-PfxCertificate -FilePath certificate/certificate.pfx -CertStoreLocation Cert:\LocalMachine\My -Password (ConvertTo-SecureString -String $WINDOWS_PFX_PASSWORD -Force -AsPlainText)
```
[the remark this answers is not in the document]
ozgunozerk: we just tried with Justin. This command did not work on my machine. The error was the referenced string being null. After removing `env:` part, it worked 🤷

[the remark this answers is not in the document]
ozgunozerk: So I tried to use the string that is being referenced by `WINDOWS_PFX_PASSWORD' directly into this command. And it worked. So it was obviously something related to `$env:WINDOWS_PFX_PASSWORD` not being able to resolve.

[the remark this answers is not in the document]
nazar-pc: It is only `null` after changes in this commit, in my commit it wasn't `null` and worked fine, in fact the whole app was built, just signature at the very end didn't work. There was no point of changing this variable usage since there was no error related to it.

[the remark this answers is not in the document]
ozgunozerk: True. For the CI, it is not needed, and now the CI is creating the same error happened on my machine with the "env:" version. 
However, isnt it weird that it is behaving differently in here? Maybe because of windows-latest, my machine is windows10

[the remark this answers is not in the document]
nazar-pc: Not sure what error you mean exactly, but please roll-back this change.

windows-latest is Windows Server 2019 at the moment on GitHub Actions, which is similar to Windows 10. Windows Server 2022 is not the default **yet**.

.github/workflows/publish.yml
```diff
@@ -6,7 +6,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        platform: [macos-latest, ubuntu-latest, windows-latest]
+        platform: [macos-latest, ubuntu-latest, windows-2019, windows-2022]
```
nazar-pc, line 9: Are you sure this will build 2 distinctly names `.msi` installers and not two with the same name that will race and only one will effectively be attached to the release?

ozgunozerk: Did this discussion resolve?
I don't have any further comments, will approve after this one :)

ImmaZoni: I spoke with the tauri team about this, and it looks like the action does not support naming the assets directly, so we will not be able to build for two windows builds right now due to this. 

The reason i added the two windows installs is due to the fact GitHub is updating `windows-latest` to windows-2022 here soon. Due to this windows-latest will not build for windows 10 here in the near future.

I am going to revert it to only build windows-latest, but we may want to figure out a solution sooner rather than later.

nazar-pc: I remember we had issues with Windows Server 2019 builds on Windows 11, so ultimately we want to have multiple builds regardless. Can you create an issue upstream and add a TODO to our action pointing to it so we can follow updates there?

ImmaZoni: Yes, i will get an issue created upstream & add the todo to the workflow.

.github/workflows/publish.yml
```diff
@@ -5,8 +5,8 @@ jobs:
   publish-tauri:
     strategy:
       fail-fast: false
-      matrix:
-        platform: [macos-latest, ubuntu-latest, windows-2019, windows-2022]
+      matrix: # Windows-2022 has been temporarily removed. We are unable to build 2x windows installs at the same name due to naming convention issues. Issue made upstream here: https://github.com/tauri-apps/tauri-action/issues/215
```
nazar-pc, line 8: Thanks! Going forward please prefer comments on a separate line though.

nazar-pc (approved):

PR #27, opened 2022-01-10
Farmer integration

ozgunozerk:
This PR addresses the following:

1. Imports farmer from our **subspace-farmer lib**, and using actual `farming/plotting` process
2. Fixes the bug in the *SetupPlot* (now, the allocated memory cannot go to negative)
3. Imported *farming* function returns the *public key*, so that frontend can use it to query the blockchain and find the farmed blocks
4. a guide for developers to test this application with the imported farming function
5. typo fixes all around

I suggest reviewing it commit by commit


src-tauri/src/main.rs
```diff
@@ -108,5 +131,120 @@ fn main() {
...
+  let rpc_server = RpcServerImpl::new(
+    farmer_metadata.record_size,
+    farmer_metadata.recorded_history_segment_size,
+    plot.clone(),
+    object_mappings.clone(),
+    subspace_codec,
+  );
+  let _stop_handle = ws_server.start(rpc_server.into_rpc())?;
+
+  info!("WS RPC server listening on {}", ws_server_addr);
```
nazar-pc, line 198: Not sure we need a WebSocket server, maybe not by default (can be added later)

src/pages/SetupPlot.vue
```diff
@@ -192,6 +198,10 @@ export default defineComponent({
       const result = await native.selectDir(this.plotDirectory).catch(console.error)
       if (result) this.plotDirectory = result
     },
+    async initFarming(path: string) {
+      const public_key = await startFarming(path)
+      return public_key
+    },
```
nazar-pc, line 204:
```suggestion
    initFarming(path: string) {
      return startFarming(path)
    },
```

No idea about Vue, but I'd really like to see return types in all these functions as well.

nazar-pc:
Also CI will need to switch to the same nightly version as in monorepo or else it will not pass

src-tauri/src/main.rs
ozgunozerk: I'm not getting this. Isn't this required? Without these lines, I'm getting this error:
```
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: Networking or low-level protocol error: Error when opening the TCP socket: Connection refused (os error 61)

Caused by:
    Error when opening the TCP socket: Connection refused (os error 61)', src/main.rs:36:14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: Networking or low-level protocol error: Error when opening the TCP socket: Connection refused (os error 61)
```

PR #30, opened 2022-01-19
Farmer lib integration

ozgunozerk:
Introduces farmer integration. Thanks to @1devNdogs.

Also improves and fixes the CI tests.

Reviewing it commit by commit is suggested


src-tauri/src/main.rs
```diff
@@ -19,24 +25,36 @@ struct DiskStats {
...
 }
 
+#[tauri::command]
+async fn farming(path: String) -> [u8; 32] {
+  let path_buf = PathBuf::from(path);
+  let node_rpc_url: String = "ws://127.0.0.1:9944".to_string();
+
+  // start farming, and return the public key of the farmer
+  let public_key = farm(path_buf, &node_rpc_url).await;
+  public_key.unwrap()
```
nazar-pc, line 35: This is less redundant:
```suggestion
  // start farming, and return the public key of the farmer
  let public_key = farm(path.into(), "ws://127.0.0.1:9944").await;
  public_key.unwrap()
```

src-tauri/src/main.rs
```diff
@@ -100,13 +119,93 @@ fn main() {
...
+
+  // wait for the farming and plotting in the background
+  tokio::spawn(async {
+    tokio::select! {
+      res = plotting_instance.wait() => if let Err(error) = res {
+        return Err(anyhow!(error))
+      },
+      res = farming_instance.wait() => if let Err(error) = res {
+        return Err(anyhow!(error))
+      },
```
nazar-pc, line 205: But these are not even printed anywhere, right?

.github/workflows/ci-tests.yml
```diff
@@ -1,31 +1,73 @@
 name: "ci tests"
-on: [push, pull_request]
+on:
+  push:
+    paths-ignore:
+      - 'docs/**'
+      - 'README.md'
+      - 'ARCHITECTURE.md'
```
nazar-pc, line 7: Never used these, but I can see them as beneficial in other repos as well :+1:

.github/workflows/ci-tests.yml
```diff
@@ -1,31 +1,73 @@
...
 
 jobs:
   test-tauri:
     strategy:
       fail-fast: false
       matrix:
-        platform: [macos-latest, ubuntu-latest, windows-latest]
+        os:
+          - ubuntu-latest
```
nazar-pc, line 20: Since we specify concrete versions, let's use `ubuntu-20.04` as well. Also maybe worth switching to macos-11 as the latest release, WDYT?

src-tauri/src/main.rs
```diff
@@ -30,30 +28,29 @@ struct DiskStats {
...
   // start farming, and return the public key of the farmer
-  let public_key = farm(path, &node_rpc_url, ws_server_listen_addr).await;
+  let public_key = farm(path_buf, &node_rpc_url).await;
   public_key.unwrap()
 }
 
 #[tauri::command]
 fn get_disk_stats(dir: String) -> DiskStats {
-  println!("{}", dir.to_string());
+  println!("{}", dir);
```
nazar-pc, line 40: Is this a debugging print?

nazar-pc (commented):
Minor comments, looks fine otherwise

.github/workflows/ci-tests.yml
ozgunozerk: Yes, the idea came to me when I remembered my experience on updating the documentation in subspace/subspace repo, and it was taking hours to pass the test, although not a single code line is changed...

.github/workflows/ci-tests.yml
ozgunozerk: I agree. If we are to follow concrete versions, changing it to `ubuntu-20.04` sounds good. However, keeping everything as `latest` is another idea. And I thought, since this will be a GUI, it may be good to know when there is a new bug emerges because of the OS's. My vote is actually on keeping everything as latest.

src-tauri/src/main.rs
ozgunozerk: currently no. They are like a placeholder at the moment, to remind my future self, it is possible to do something with them, or maybe get rid of them :)

nazar-pc: TODO would make it clean :wink:

src-tauri/src/main.rs
ozgunozerk: yes, printing to the CLI. It was there already, I did not bother changing it. I will change it to `log::info`, don't want to remove that, may become handy :)

src-tauri/Cargo.toml
```diff
@@ -6,16 +6,23 @@ authors = [ "you" ]
...
 
 [build-dependencies]
-tauri-build = { version = "1.0.0-beta.2" }
+tauri-build = { version = "1.0.0-beta.4" }
 
 [dependencies]
+anyhow = "1.0.44"
+dirs = "4.0.0"
+fs2 = "0.4.3"
+jsonrpsee = { version = "0.6.1", features = ["client", "macros", "server"] }
```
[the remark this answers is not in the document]
ozgunozerk: We are using `anyhow`, deleted the others, thanks!

nazar-pc (approved):

PR #31, opened 2022-01-25
formatting (no new code)

ozgunozerk:
There is no new functionality introduced.
- Vue, Js and Ts files are formatted automatically by the editor.
- Rust files are indented with 4 spaces as the convention instead of 2.
- rustfmt file is removed


src/components/SaveKeys.vue
```diff
@@ -2,69 +2,89 @@
 .full-width
   .row.justify-center.q-gutter-lg
     .col-auto
-      q-icon(color="blue-3" name="vpn_key" size="80px")
+      q-icon(color="blue-3", name="vpn_key", size="80px")
```
nazar-pc, line 5: These comas are super weird in this context, this is not how HTML (from which pug syntax is derived) works

nazar-pc (commented):
Please get rid of comas in pug templates, otherwise looks fine to me

src/components/SaveKeys.vue
ozgunozerk: I figured out these are called `attribute separators`. The default was **comma** in prettier (vetur is using prettier for formatting pug fields). So a new file is introduced called ".prettierrc", along with these commas, all attribute separators are deleted. Is it fine now?

nazar-pc (approved):
Fine with me overall, next step is to enforce these in CI :slightly_smiling_face:

PR #32, opened 2022-01-26
mnemonic support added to desktop app

ozgunozerk:
This PR extracts the mnemonic from the farmer lib from the tauri backend, and uses it in the desktop app

### DIFF
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -8,7 +8,7 @@ mod windows;
 
 use anyhow::{anyhow, Result};
 use bip39::{Language, Mnemonic};
-use log::{info, debug};
+use log::{debug, info};
 use serde::Serialize;
 use std::path::PathBuf;
 use subspace_farmer::{
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -7,6 +7,7 @@
 mod windows;
 
 use anyhow::{anyhow, Result};
+use bip39::{Language, Mnemonic};
 use log::info;
 use serde::Serialize;
 use std::path::PathBuf;
@@ -21,21 +22,30 @@ use tauri::{
 };
 
 #[derive(Serialize)]
+struct FarmerIdentity {
+    public_key: [u8; 32],
+    mnemonic: String,
+}
+
+#[derive(Serialize)]
 struct DiskStats {
     free_bytes: u64,
     total_bytes: u64,
 }
 
 #[tauri::command]
-async fn farming(path: String) -> [u8; 32] {
+async fn farming(path: String) -> FarmerIdentity {
+    let path_buf = PathBuf::from(path);
+    let node_rpc_url: String = "ws://127.0.0.1:9944".to_string();
+
     // start farming, and return the public key of the farmer
-    let public_key = farm(path.into(), "ws://127.0.0.1:9944").await;
-    public_key.unwrap()
+    let farmer_identity = farm(path_buf, &node_rpc_url).await;
+    farmer_identity.unwrap()
 }
 
 #[tauri::command]
 fn get_disk_stats(dir: String) -> DiskStats {
-    info!("{}", dir);
+    println!("{}", dir);
     let free: u64 = fs2::available_space(&dir).expect("error");
     let total: u64 = fs2::total_space(&dir).expect("error");
 
@@ -135,7 +145,7 @@ async fn main() -> Result<()> {
 pub(crate) async fn farm(
     base_directory: PathBuf,
     node_rpc_url: &str,
-) -> Result<[u8; 32], anyhow::Error> {
+) -> Result<FarmerIdentity, anyhow::Error> {
     // TODO: This doesn't account for the fact that node can
     // have a completely different history to what farmer expects
     info!("Opening plot");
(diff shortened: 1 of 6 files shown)


src-tauri/src/main.rs
```diff
@@ -20,22 +21,27 @@ use tauri::{
...
-    public_key.unwrap()
+async fn farming(path: String) -> FarmerIdentity {
+    let farmer_identity = farm(path.into(), "ws://127.0.0.1:9944").await;
+    farmer_identity.unwrap()
 }
 
 #[tauri::command]
 fn get_disk_stats(dir: String) -> DiskStats {
-    info!("{}", dir);
+    println!("{}", dir);
```
nazar-pc, line 44: Did you actually mean to do this? Looks like a debug message to me at best.

src-tauri/src/main.rs
```diff
@@ -195,15 +201,21 @@ pub(crate) async fn farm(
...
         Ok(())
     });
 
-    Ok(identity.public_key().to_bytes())
+    Ok(FarmerIdentity {
+        public_key: identity.public_key().to_bytes(),
+        mnemonic: Mnemonic::from_entropy(identity.entropy(), Language::English)
+            .unwrap()
+            .phrase()
+            .parse()?,
```
nazar-pc, line 219: Parsing `&str` to get `String` is non-idiomatic and I think in this case not needed:
```suggestion
        mnemonic: Mnemonic::from_entropy(identity.entropy(), Language::English)
            .unwrap()
            .into_phrase(),
```

src-tauri/src/main.rs
```diff
@@ -195,15 +201,21 @@ pub(crate) async fn farm(
...
-          },
-          res = farming_instance.wait() => if let Err(error) = res {
-            return Err(anyhow!(error)) // TODO: connect this error to frontend, or log it
-          },
+            res = plotting_instance.wait() => if let Err(error) = res {
+                return Err(anyhow!(error))
+            },
+            res = farming_instance.wait() => if let Err(error) = res {
+                return Err(anyhow!(error))
+            },
```
nazar-pc, line 209: Why were TODOs removed?

src/lib/client.ts
```diff
@@ -63,6 +63,11 @@ export interface ClientData {
   farming: ClientFarming
 }
 
+export interface ClientIdentity {
+  public_key: string,
```
nazar-pc, line 67: It is non-idiomatic to use `snake_case` in TypeScript, I would suggest to add `#[serde(rename_all = "snake_case")]` to the struct on backend.

src/lib/client.ts
```diff
@@ -203,6 +208,8 @@ export class Client {
   }
 }
 
-export function startFarming(path: string): Promise<string> {
-  return (tauri.invoke('farming', { path })).then(value => value as string)
+export async function startFarming(path: string): Promise<ClientIdentity> {
+  const result = await (tauri.invoke('farming', { path })).then(value => value as ClientIdentity)
+  const farmer_identity: ClientIdentity = { public_key: result.public_key, mnemonic: result.mnemonic }
+  return farmer_identity
 }
```
nazar-pc, line 215: This should be sufficient I think:
```suggestion
export function startFarming(path: string): Promise<ClientIdentity> {
  return tauri.invoke('farming', { path }));
}
```

src/pages/SetupPlot.vue
```diff
@@ -161,8 +161,7 @@ export default defineComponent({
     return {
       revealKey: false,
       userConfirm: false,
-      generatedPk:
-        "98da4ef21b864d2cc526dbdb2a120bd2874c36c9d0a1fb7f8c63d7f7a8b41de8f",
+      //generatedPk: "98da4ef21b864d2cc526dbdb2a120bd2874c36c9d0a1fb7f8c63d7f7a8b41de8f",
```
nazar-pc, line 164: Time to remove?

src/pages/SetupPlot.vue
```diff
@@ -262,9 +261,10 @@ export default defineComponent({
       })
       if (this.defaultPath != this.plotDirectory)
         await native.createDir(this.plotDirectory)
-      const public_key = await startFarming(this.plotDirectory)
+      const farmer_identity = await startFarming(this.plotDirectory)
       // TODO: find a way to store and retrieve the public key from client.ts.
-      LocalStorage.set("farmerPublicKey", public_key)
+      LocalStorage.set("farmerPublicKey", farmer_identity.public_key)
+      LocalStorage.set("mnemonic", farmer_identity.mnemonic)
```
nazar-pc, line 267: Storing mnemonic here this way is not a great idea, but can be dealt with later as things will change anyway.

src-tauri/src/main.rs
ozgunozerk: Forgot to merge main into this branch... Fixing similar ones as well now

src/lib/client.ts
ozgunozerk: Changing this (line 67) to `publicKey` and changing where it is used:
[src/pages](https://github.com/subspace/subspace-desktop/pull/32/files#diff-e879b9701cf110ae7a5636e48ba2803b2f6a8b808b3539e2937607f7a0200d11R265)
`farmerIdentity.public_key` into `farmerIdentity.publicKey` breaks some functionality. 
The desktop app cannot retrieve the solutions that have been found by the farmer and thus displays **0 solutions found**, although it finds solutions...

Don't know the cause yet. Also tried to change other `public_key` namings to `publicKey`, still did not work. I believe it has to match with tauri-side variable naming. Not sure though

src/lib/client.ts
```diff
@@ -203,6 +208,6 @@ export class Client {
   }
 }
 
-export function startFarming(path: string): Promise<string> {
-  return (tauri.invoke('farming', { path })).then(value => value as string)
+export async function startFarming(path: string): Promise<ClientIdentity> {
```
nazar-pc, line 211: This will work, but function doesn't need to be `async` in this case since there is no `await` in it.

nazar-pc (approved):

src/lib/client.ts
ozgunozerk: alright, will include in the next pr :)

PR #39, opened 2022-02-15
Front end - Client, api and storage updates

1devNdogs:
# Sumary.

src/lib/client.ts
- Move interfaces from to types.
- Update class to work from different pages.
- Fix load, clear and store farmed blocks.
- Add block reward on farmed block.

src/pages/Index.vue
- On app mounted, check if plot and account exist and redirect to dashboard page.

src/pages/Dashboard.vue
- Updates from Client class.

src/lib/util.ts
- LocalStorage.clear() added on reset app.

src/components/farmedList.vue
- Fix table and css to fit new content.


src/lib/client.ts
```diff
@@ -152,59 +47,121 @@ export class Client {
...
-        },
-        SubPreDigest: {
-          slot: 'u64',
-          solution: 'Solution'
+
+  // If the app is started for the first time, the client will be started from PlottingProgress page.
+  // In this case, farmerPublicKey and mnemonic must be stored and clear stored farmedBlocks if exist (clearStoredBlocks).
+  // -
+  // If the app is started for the second time, the client will be started from Dashboard page.
+  // In this case, farmerPublicKey and mnemonic already exist and just need to load the stored farmedBlocks (loadStoredBlocks).
```
ozgunozerk, line 104: When the app is started for the first time, it should start from the password setup page, then proceed into plotSetup -> plottingProgress -> Modals for extra info and private key -> Dashboard

if it is started for the second time, it should start directly from plotting progress (because probably it needs to plot some additional pieces for the initial plotting, since the farmer was offline for a bit. 
So in plotting progress, maybe it will start from 90% progress, and will plot for the remaining 10%. 
Another difference would be, we shouldn't display the info Modals for the second time.

However, I think this version is fine at the moment. We will add/update these functionalities in a later PR.

src/pages/Index.vue
```diff
@@ -23,12 +23,23 @@ q-page(padding)
...
+import { startFarming } from "src/lib/client"
 const lang = global.data.loc.text.index
 
 export default defineComponent({
   data() {
     return { lang }
   },
+  async mounted() {
+    const configData = await config.read()
+    // If the app star and config with plot and account exists, start the farmer
```
ozgunozerk, line 36: as I explained in the above comment, I believe it is fine for now. But maybe we can add a TODO here to redirect into plottingProgress page, and not to show information again.

Or, we can make the mnemonic modal a separate one, and will be before plottingProgress page. That way, we can just redirect to plottingProgress page here, without additional logic

src/pages/SetupPlot.vue
```diff
@@ -249,10 +250,9 @@ export default defineComponent({
       })
       if (this.defaultPath != this.plotDirectory)
         await native.createDir(this.plotDirectory)
-      const farmerIdentity = await startFarming(this.plotDirectory)
-      // TODO: find a way to store and retrieve the public key from client.ts.
-      LocalStorage.set("farmerPublicKey", farmerIdentity.publicKey)
-      LocalStorage.set("mnemonic", farmerIdentity.mnemonic)
+      const { publicKey, mnemonic } = await startFarming(this.plotDirectory)
+      if(publicKey && mnemonic)
+        await this.client.init(publicKey, mnemonic)
```
ozgunozerk, line 255: maybe we can display the mnemonic here directly in a modal, without storing it. 
This will resolve the potential security vulnerability we have (storing a piece of sensitive information).

ozgunozerk:
These are really great changes imo. 
I have a few questions/comments that I wrote above, but they can be implemented in the later PR's. 
Will this be squash-merged? Because commits seem a bit messy :D
If we want to include the commits in the history, I might suggest a rebase

src/lib/client.ts
ImmaZoni: I would think it would be better to have a plotting indicator on the main farming page, once the initial "batch" of plotting is completed. just something in the corner next to network status, could just be "plotting status" with a light, and a tooltip. 

But once again, just thoughts for a later version.

1devNdogs:
> These are really great changes imo. I have a few questions/comments that I wrote above, but they can be implemented in the later PR's. Will this be squash-merged? Because commits seem a bit messy :D If we want to include the commits in the history, I might suggest a rebase

I can squash commits, next branch PR will consider some of these comments, will add TODO also.

src/pages/Index.vue
1devNdogs: Wath if. In case the farmer got 10 days disconnected and the user re open the app, we can keep going to dashboard and update the Plot Status Label with a loader and % progress. In the next branch we already have the data to show it. WDYT?

src/lib/client.ts
1devNdogs: Just as a comment, this changes does not affect the way the pages are loaded or rendered, changes only the way the client class and polkadot api is initialized and loaded.  

https://github.com/subspace/subspace-desktop/pull/39#discussion_r807440890

ozgunozerk (approved):
Great work and progress here. 
If we include the comments and TODO's as the next PR, I think we are good to go for this one :)

PR #46, opened 2022-02-16
plotting progress

ozgunozerk:
@1devNdogs 
Leo I remember you already implemented how to use this new functionality and it was working (we were working on this branch). 
You can commit your changes too in this PR, and we can review it later together (and will also assign Nazar as a reviewer after we finalize this PR).

- [x] We also have to change the default 10GB allocated space with the actual data to be plotted, queried from the blockchain. (will do later in an upcoming PR)


src/pages/PlottingProgress.vue
```diff
@@ -214,21 +217,62 @@ export default defineComponent({
...
+      
+      this.plottingData.status = lang.startingFarmer
+      // After local node is fully Synced, the farmer will be able to actualy plot and farm. 
+      const { publicKey, mnemonic } = await startFarming(this.plotDirectory.replace("/subspace.plot", ""))
+ 
+      // When farmer start, temporal mnemonic storage, when user accept on modal that mnemonic was backed up. We remove it from local storage
+      // Check SaveKeys.vue.
+      if (publicKey && mnemonic) {
+        // TODO: Fully remove LocalStorage. Find a secure way to pass mnemonic to SaveKeys modal.
+        LocalStorage.set("mnemonic", mnemonic)
```
ozgunozerk, line 249: I couldn't see how mnemonic is removed from LocalStorage after here? 
I saw a reset on mnemonic on here: https://github.com/subspace/subspace-desktop/pull/46/files#diff-653e0382a5f2a0bcb258b7abb55fe7ac7e6e00e6e9295618f526671a54d11cc5R66

But I believe that code will be executed before here?

src-tauri/src/main.rs
```diff
@@ -151,8 +175,23 @@ pub(crate) async fn farm(
 
         move || Plot::open_or_create(&base_directory)
     });
+
     let plot = plot_fut.await.unwrap()?;
 
+    let mut _handler_guard = HANDLER
```
nazar-pc, line 181: I don't understand why you have `let mut _handler_guard =` and why you need to store it at all. Maybe you can just call `.detach()` on handler instead?

src-tauri/src/main.rs
```diff
@@ -6,11 +6,17 @@
 #[cfg(target_os = "windows")]
 mod windows;
 
+#[macro_use]
+extern crate lazy_static;
```
nazar-pc, line 10: `extern crate` is 2015 Rust style, use `use` instead.

1devNdogs (approved):

src-tauri/src/main.rs
nazar-pc: Why do you need to store handler at all?

ozgunozerk: I don't. It was there because of the old trials. I don't store it, this is changed with the last commit

ozgunozerk: I thought you were referring to `let mut _handler_guard`. 
So I need a `HANDLER` to be able to keep track of updates on the plotting pieces. 
This needs to be initiated in `farm` function. However, `farm` function returns quickly, and because of that the `HANDLER` needs to be a global variable to outlive the farmer. At least, that was the solution I came up with.

nazar-pc: I was referring both to that (because it wasn't even a guard) and the fact that you store anything in the first place. I also mentioned method that you can call instead of storing it and achieve the same result with less dependencies and less code.

ozgunozerk: I think it is fine now :)

nazar-pc (approved):

PR #54, opened 2022-02-19
Publish update

ozgunozerk:
This updates the `publish.yml`. 
Right now, `publish.yml` is outdated, so we need to update it for releasing the desktop-app.

We also don't want regular CI jobs to trigger when `yml` files are edited. The reason is, yaml file changes only take effect when they are merged into main. So running the current tests when a `yml` file changes provides no extra information to us, just wastes time.

pre-release is set to true, I will change it back to `false` when we successfully integrate the node, and the desktop-app can be a standalone farming application.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -19,7 +19,7 @@ jobs:
       - name: setup node
         uses: actions/setup-node@v1
         with:
-          node-version: 14
+          node-version: 16
 
       - name: install Rust nightly
         uses: actions-rs/toolchain@v1
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -6,19 +6,34 @@ jobs:
     strategy:
       fail-fast: false
       matrix: # Windows-2022 has been temporarily removed. We are unable to build 2x windows installs at the same name due to naming convention issues. Issue made upstream here: https://github.com/tauri-apps/tauri-action/issues/215
-        platform: [macos-latest, ubuntu-latest, windows-2019]
+        os:
+          - ubuntu-latest
+          - macos-latest
+          - windows-2019
+          - windows-2022 # test again to see if it will fail, comment out this again if necessary
 
     runs-on: ${{ matrix.platform }}
     steps:
-      - uses: actions/checkout@v2
+      - name: git checkout
+        uses: actions/checkout@v2
+
       - name: setup node
         uses: actions/setup-node@v1
         with:
           node-version: 14
-      - name: install Rust stable
+
+      - name: install Rust nightly
         uses: actions-rs/toolchain@v1
         with:
-          toolchain: stable
+          toolchain: nightly-2021-12-14
+          override: true
+          components: rustfmt, clippy
+
+        # Workaround to resolve link error with C:\msys64\mingw64\bin\libclang.dll
+      - name: Remove msys64
+        run: Remove-Item -LiteralPath "C:\msys64\" -Force -Recurse
+        if: runner.os == 'Windows'
+
       - name: install webkit2gtk (ubuntu only)
         if: matrix.platform == 'ubuntu-latest'
         run: |
@@ -26,6 +41,7 @@ jobs:
           sudo apt-get install -y --no-install-recommends libwebkit2gtk-4.0-dev libappindicator3-dev0
 
       # Imports PFX Certificate into keystore, allows Tauri to then sign the exe without being passed the private key.
+      # This normally should not be required on CI-Tests, but due to the way that tauri has its code signing configuration for windows it is required that we sign or the build will fail.
       - name: import windows certificate
         if: runner.os == 'Windows'
         env:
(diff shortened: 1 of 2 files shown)


.github/workflows/publish.yml
```diff
@@ -6,26 +6,42 @@ jobs:
     strategy:
       fail-fast: false
       matrix: # Windows-2022 has been temporarily removed. We are unable to build 2x windows installs at the same name due to naming convention issues. Issue made upstream here: https://github.com/tauri-apps/tauri-action/issues/215
-        platform: [macos-latest, ubuntu-latest, windows-2019]
+        os:
+          - ubuntu-latest
+          - macos-latest
+          - windows-2019
+          - windows-2022 # test again to see if it will fail, comment out this again if necessary
```
nazar-pc, line 13: We already discussed that this will make 2 Windows builds competing with each other and only one will win, but we don't know which one.

.github/workflows/publish.yml
```diff
@@ -6,26 +6,42 @@ jobs:
...
         uses: actions/setup-node@v1
         with:
           node-version: 14
-      - name: install Rust stable
+
+      - name: install Rust nightly
         uses: actions-rs/toolchain@v1
         with:
-          toolchain: stable
+          toolchain: nightly-2021-12-14
```
nazar-pc, line 28: This is an old one, with latest node we should really switch to just `nightly` everywhere.

.github/workflows/publish.yml
```diff
@@ -6,26 +6,42 @@ jobs:
...
+        if: runner.os == 'Windows'
+
       - name: install webkit2gtk (ubuntu only)
         if: matrix.platform == 'ubuntu-latest'
         run: |
           sudo apt-get update
           sudo apt-get install -y --no-install-recommends libwebkit2gtk-4.0-dev libappindicator3-dev0
 
       # Imports PFX Certificate into keystore, allows Tauri to then sign the exe without being passed the private key.
+      # This normally should not be required on CI-Tests, but due to the way that tauri has its code signing configuration for windows it is required that we sign or the build will fail.
```
nazar-pc, line 44: This message is incorrectly copy-pasted from another place, but it is confusing here because this is a publishing action, it has to import certificate to sign release later.

.github/workflows/publish.yml
```diff
@@ -37,10 +53,16 @@ jobs:
           certutil -decode certificate/tempCert.txt certificate/certificate.pfx
           Remove-Item –path certificate -include tempCert.txt
           Import-PfxCertificate -FilePath certificate/certificate.pfx -CertStoreLocation Cert:\CurrentUser\My -Password (ConvertTo-SecureString -String $env:WINDOWS_PFX_PASSWORD -Force -AsPlainText)
+
       - name: install app dependencies and build it
         run: yarn && yarn build
-      - uses: tauri-apps/tauri-action@v0
         env:
+          MACOSX_DEPLOYMENT_TARGET: "10.13"
```
nazar-pc, line 60: Why is this necessary?

.github/workflows/publish.yml
```diff
@@ -37,10 +53,16 @@ jobs:
...
+
+      - name: tauri run
+        uses: tauri-apps/tauri-action@v0
+        env:
+          MACOSX_DEPLOYMENT_TARGET: "10.13"
```
nazar-pc, line 65: Ad why is this necessary here again?

.github/workflows/publish.yml
```diff
@@ -53,4 +75,6 @@ jobs:
           releaseName: "App v__VERSION__"
           body: "See the assets to download this version and install."
           draft: true
-          prerelease: true
+          prerelease: false
+
+    # rust fmt and clippy actions are not included, since we should only release a version that is passed the tests
```
nazar-pc, line 80: Why is this explanation even here? Obviously we don't need them in publish action, because its job is to publish, not to test...

nazar-pc (changes requested):
> yaml file changes only take effect when they are merged into main

I don't think so. Adding and/or removing operating systems definitely takes effect immediately.

Just like you can see in this PR: CI isn't running at all.

.github/workflows/publish.yml
ozgunozerk: yeah, since it was some time ago, I thought maybe something is changed 
I went to the issue link, but there was no update :(

.github/workflows/publish.yml
ozgunozerk: oh yeah, my bad, applying now

.github/workflows/publish.yml
ozgunozerk: We had this in the CI-tests, so I thought it is probably helpful to have. Don't have strong background on the necessity. 
I remember we were doing some trial/error due to macos version compatibility. Probably a remnant of that.

Do you want us to get rid of this?

.github/workflows/publish.yml
ozgunozerk: > > yaml file changes only take effect when they are merged into main
> 
> I don't think so. Adding and/or removing operating systems definitely takes effect immediately.
> 
> Just like you can see in this PR: CI isn't running at all.

True... Reverting the `yml` ignore changes

.github/workflows/publish.yml
ozgunozerk: So, we should have 2 publish actions then? 
1 for windows2019, macos and linux
1 for windows2022
?

.github/workflows/publish.yml
nazar-pc: Both in tests and here. Or explain why it is necessary with link to upstream issue or whatnot.

.github/workflows/publish.yml
nazar-pc: It builds executable with the same name regardless of number of actions. An alternative would be to build executable without uploading it, renaming it and uploading renamed version (separate for Windows 10 and 11). This is all in case it is even needed in the first place (namely if desktop app doesn't run on both otherwise).

ozgunozerk:
Macos builded and ran without failing after removing the `macosx_deployment_target`. 
Windows tests does not hang and fail anymore on the `yarn build` step. 
I think we are good here! 🥳

nazar-pc (approved):

PR #55, opened 2022-02-20
Publish ci fix

ozgunozerk:
After the update on `subspace-network`, tests are passing!

Also a small fix to `publish.yml` that I missed on the previous PR.


nazar-pc (dismissed):
I would strongly recommend to specify a revision in addition to the git repository, makes it easier to upgrade/downgrade and review.

nazar-pc (approved):

PR #57, opened 2022-02-22
Update icon padding

### LINKED ISSUE #52: MacOS dock icon is a little bit too big
Although it is a bit too big, it doesn't look ugly, resolution is quite nice, and it looks outstanding compared to other apps. This is for sure not a priority, but we can fix it sometime later.

image

1devNdogs:
Fixes #52

@ozgunozerk can you check bundles on mac and windows before PR?


ozgunozerk:
macOS icon is still a bit big, I believe the linked issue is not resolved yet. 
I will try on windows too

ozgunozerk:
This is the dock (without hover): 
image

This is the magnification effect on the dock when hovered:
image

ozgunozerk:
On windows, there are 2 problems:
report_windows
Orange and green circled icons resolutions are very low. It was better before this PR imo.

PR #69, opened 2022-03-04
Node integration without CLI

nazar-pc:
This removes both `subspace-node` and `sc-cli` from dependencies since we want to run subspace node as a library, not as CLI application.

Have not tested this, but I think it should work. Chain spec might need to be modified to include correct bootstrap node, `FORCE_AUTHORING` environment variable can be used to force authoring. I also had to do a few minor tweaks.

Makes https://github.com/subspace/subspace/pull/274 unnecessary.

### DIFF
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -172,16 +172,21 @@ async fn init_node(base_directory: PathBuf) -> Result<FarmerIdentity> {
         )
         .map_err(anyhow::Error::msg)?;
 
-    let mut full_client =
-        Handle::current().block_on(node::create_full_client(chain_spec, base_directory))?;
+    let handle = Handle::current();
 
-    // TODO: Make this interruptable if needed
-    tokio::spawn(async move {
-        if let Err(error) = full_client.task_manager.future().await {
-            error!("Task manager exited with error: {error}");
-        } else {
-            error!("Task manager exited without error");
-        }
+    std::thread::spawn(move || {
+        let mut full_client = handle
+            .block_on(node::create_full_client(chain_spec, base_directory))
+            .unwrap();
+
+        // TODO: Make this interruptable if needed
+        handle.spawn(async move {
+            if let Err(error) = full_client.task_manager.future().await {
+                error!("Task manager exited with error: {error}");
+            } else {
+                error!("Task manager exited without error");
+            }
+        });
     });
 
     Ok(FarmerIdentity {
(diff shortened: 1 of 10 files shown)


ozgunozerk:
I did an investigation on the generated `config`. Printed it in a file, and compared it to the `config` we generate from the **subspace/subspace/crates/node**. 

There are some significant differences, don't know exactly which ones are critical, can only guess. 

Here is the report: the differences are noted as comments
[config-desktop.txt](https://github.com/subspace/subspace-desktop/files/8184430/config-desktop.txt)

ozgunozerk:
We have worked with Leo, was experimenting on another branch based on your changes, and made it work with the current setting. We will be creating another PR from that working branch

src-tauri/src/main.rs
```diff
@@ -172,16 +172,21 @@ async fn init_node(base_directory: PathBuf) -> Result<FarmerIdentity> {
...
-    // TODO: Make this interruptable if needed
-    tokio::spawn(async move {
-        if let Err(error) = full_client.task_manager.future().await {
-            error!("Task manager exited with error: {error}");
-        } else {
-            error!("Task manager exited without error");
-        }
+    let handle = Handle::current();
+
+    std::thread::spawn(move || {
```
nazar-pc, line 177: Why? The idea here was to check for the error before returning from the async function. If blocking is undesired, blocking task needs to be spawned and waited for.

nazar-pc (commented):
> Here is the report: the differences are noted as comments

Thanks, attaching proper diff would be better though. We should probably change some of them, but not all. I'm surprised so many values have non-default values, not immediately clear where it comes from.

nazar-pc:
Should be fine now

ozgunozerk:
From our trials with Leo (he started a new chain from up to date node), we were able to connect, sync to the node, archive and farm new blocks.

Without `state_pruning: PruningMode::ArchiveAll`, we cannot farm new blocks, it throws the error `state already discarded for block #0`.

 With `Default::default()`, it is creating the config with `Constrained`.

nazar-pc:
So can we merge this in or what?

ozgunozerk:
You did not reply to my above comments, I was waiting for them. Do we want to keep `generate_name()`? Or use the public key? Also I assume using `InMemory` for `keystore` is not important?

nazar-pc:
It doesn't matter what name is used for node except when it is shown on telemetry page AFAIK, so generated name is fine for now.

We have no keystore used in our node, so it doesn't matter at all.

ozgunozerk (approved):

PR #70, opened 2022-03-08
Switch CI to Windows 2022

nazar-pc:
Seems to work fine both on Windows 10 and WIndows 11, so we're switching monorepo to only build under Windows 2022, let's make the same here unless you are aware of any issues this may cause.

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -15,7 +15,6 @@ jobs:
         os:
           - ubuntu-latest
           - macos-latest
-          - windows-2019
           - windows-2022
 
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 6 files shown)


ImmaZoni (dismissed):

nazar-pc:
Now yarn fails with 502 :disappointed: 
I remember it working better in the past.

nazar-pc:
Have you seen this error before?

ozgunozerk:
Very interesting. 
See the `finished` section here, look at the path of the bundle (this is failing build): https://github.com/subspace/subspace-desktop/runs/5458464406?check_suite_focus=true

And see the same section here, look at the path of the bundle (this is the successful build from past): https://github.com/subspace/subspace-desktop/runs/5439464934?check_suite_focus=true

In the previous CI runs, there was no `expected artifact paths` requirement for macos, ubuntu or windows. This is something new, and it broke windows tests. Don't know exactly why.

nazar-pc:
Reported as https://github.com/tauri-apps/tauri-action/issues/240

ozgunozerk (approved):

Issue #61, opened 2022-02-25
Integrate Node to Farmer


ozgunozerk:
- [ ] Run a plain node from the subspace desktop app
- [ ] Pass the necessary parameters to node (for online farming)
- [ ] Pass the chain_spec to node (for online farming)
- [ ] Expose the `start_node` function to frontend
- [ ] Integrate the `start_node` function to the current workflow in the Frontend side

Issue #48, opened 2022-02-17
Initial Plotting - Slow archiving proccess.


1devNdogs:
Running the desktop farmer, the farmer takes around 1-2 second to archive each segment, this result in the user stuck a lot of time in the PlottingProgress page.
(Connected to farmnet this takes like 1-2 hours to fully create the plot from the desktop app)

After testing, running a farmer from the command line take very much less time to archive. 
(on farmnet this takes minutes to fully create the plot from a regular farmer from cli)

ozgunozerk:
cuda was not a default feature for the farmer. Adding this feature to the subspace-desktop. You should see increased speed :)

1devNdogs:
> cuda was not a default feature for the farmer. Adding this feature to the subspace-desktop. You should see increased speed :)

@ozgunozerk. 

As @nazar-pc said in the last meet, we need to add `[profile.dev.package]` section on Cargo.toml, the same way as `subspace/subspace` [Cargo.toml](https://github.com/subspace/subspace/blob/c50248790d86e5cac5174145969a5f7cb2d16049/Cargo.toml#L43-L98
)  I just removed a few, not present dependencies and the `archiving speed` is working as expected. 

@nazar-pc  I have not included [`profile.production`] and [`profile.release`]. Think we don't need them here, is that true?

https://github.com/subspace/subspace-desktop/blob/farmer-archiving-speed/src-tauri/Cargo.toml

nazar-pc:
CUDA will not increase it significantly with current architecture, it will likely make it slower.

> Think we don't need them here, is that true?

It would be better to have them and to build production app with production, not just release profile (it is even slower, but results in more optimized binaries). See https://github.com/subspace/subspace/pull/263 for examples of how it is used.

ozgunozerk:
Do we still have this? I believe it is resolved.

1devNdogs:
Not in main yet, but on branch [front-end-node-integration](https://github.com/subspace/subspace-desktop/tree/front-end-node-integration) got fixed : https://github.com/subspace/subspace-desktop/blob/f237b50b72ca2cb864035196ade7a8fb7dd83cc3/src-tauri/Cargo.toml#L53

ozgunozerk:
Ok great, so after we merge that, this should be closed as well :)

PR #80, opened 2022-03-10
Node integration

ozgunozerk:
Kind of a big PR. 

Introduces node integration to subspace desktop. Turning it into a standalone app.

Also solves #79, #48 and #61.

Question: should we bump the version into 0.2.0 with this PR (I can do a quick rebase), or should that be another PR? 
I don't know if there is a best practice for that.


nazar-pc (approved):
Looks fine to me, I'm just wondering maybe we should change bootstrap node to point to the (currently) inactive farmnet, such that it works even before we make a switch.

ozgunozerk:
We actually discussed that in the meeting. We will be making the switch pretty soon, that's why we decided to make this change and be ready for it. But I'm fine with both strategies :)
I believe @1devNdogs and @ImmaZoni are more knowledgeable on this issue than me

nazar-pc:
Well, it will be broken before the switch though, meaning we can't test the final builds. It also means that when we do the switch next time, we'll break the previously produced builds instead of giving them some more time to switch. I think setting more specific endpoint is actually beneficial because of that.

1devNdogs:
> Well, it will be broken before the switch though, meaning we can't test the final builds. 

Oh, I hadn't considered that. I have updated values to target the inactive network on `aries-farm-rpc-b`

nazar-pc (dismissed):

nazar-pc:
It would be nice to add advanced settings in the future where we can customize bootstrap nodes and other settings.

1devNdogs (dismissed):

ozgunozerk:
So I guess we will wait for the switch to happen, then update the version and bootnodes in the code, and have a release. 
In the meantime, we can merge this PR. 
I have squashed the last tweak commit, and enabled auto-merge.

nazar-pc (approved):
We don't need to update bootnodes after, they'll continue to work and for next release of node we'll need new desktop release anyway.

PR #100, opened 2022-03-18
Entitlements revamp

ozgunozerk:
This is the crux of the problem for macos signing.

**Please don't feel lazy and read all the below, they are important.**

### What we got right:
- Signing: checked with the following command `codesign -vvv --deep --strict ./subspace-desktop.app`
- Notarization: checked with the following command `spctl -vvv --assess --type exec ./subspace-desktop.app`

### What is wrong:
The app is crashing during `node sync` or `plotting` phases. If our signature was not valid, the app wouldn't open in the first place. 
After some research, I found out there are also runtime related things for the signature (setting the permissions for the app).
So my educated guess is on these runtime related permissions. 

I found out there are 2 keys related to network operations. See below:
```
    <key>com.apple.security.network.server</key> <!-- required for incoming connections -->
    <true/>
    <key>com.apple.security.network.client</key> <!-- required for outgoing coonnections -->
    <true/>
```

I already added these. But there are more stuff that we may require. Hence, it was important for you to read above, and we can discuss the below:

image

their keys would correspond to:
```
    <key>com.apple.security.cs.allow-unsigned-executable-memory</key>
    <true/>
    <key>com.apple.security.cs.disable-library-validation</key>
    <true/>
    <key>com.apple.security.cs.allow-dyld-environment-variables</key>
    <true/>
    <key>com.apple.security.cs.allow-jit</key>
    <true/>
```

So, which one of the above do you think we need (or do we need anything at all)? 
I don't know much about them, currently researching. Any help is appreciated!

### DIFF
--- a/src-tauri/entitlements.plist
+++ b/src-tauri/entitlements.plist
@@ -2,7 +2,11 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
-    <key>com.apple.security.files.user-selected.read-write</key>
+    <key>com.apple.security.network.server</key> <!-- required for incoming connections -->
+    <true/>
+    <key>com.apple.security.network.client</key> <!-- required for outgoing coonnections -->
+    <true/>
+    <key>com.apple.security.files.user-selected.read-write</key> <!-- required for read/write access to files -->
     <true/>
 </dict>
 </plist>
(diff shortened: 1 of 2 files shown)


nazar-pc (commented):
I don't think we need other entitlements, but why are the changes to UI included here?

ozgunozerk:
Yes my bad, rebasing now. Apperantly I started this from the nodesync-farmer branch.

nazar-pc:
There is still many new lines removed in this PR for some reason

ozgunozerk:
I'm now testing this. If it succeeds, I will open a clean PR, so we can merge and test it together.

Issue #75, opened 2022-03-10
Make `start on boot` option the default one


ozgunozerk:

ozgunozerk:
After some debugging, I found out that `enable` function for `macAL` is not working. It is under the file `autoLauncher.ts`, in block starting with `const macAL = {`

ozgunozerk:
Investigated for mac. It is working, but the source code lost some flexibility and I don't like it. Please see the below two links:

1. https://github.com/subspace/subspace-desktop/commit/3a579872ab3233ccc4c5e6375adcd9540cf3156d#diff-b64a4ad147b7c8b2102e414ad2a6a5f453e6dff8e468b75a3e6d53768bd417c1R74
2. https://github.com/subspace/subspace-desktop/commit/3a579872ab3233ccc4c5e6375adcd9540cf3156d#diff-b64a4ad147b7c8b2102e414ad2a6a5f453e6dff8e468b75a3e6d53768bd417c1R80

I've spent around 1-hour trying alternate versions. If we cannot find a better way, we can still use the above solution, it works.

1devNdogs:
@ozgunozerk , did you test this on `dev` mode or run the app from `build`?

I'm asking because, on Linux, running on `tauri dev` the **appName** was always returning the string `app`, getting an autostart file called **app.desktop**  and running from build I was getting **subspace-desktop.desktop** as expected and defined on main branch:
https://github.com/subspace/subspace-desktop/blob/51eb2ca4cbb7519b08cefbd49e02025356bc80d3/src/lib/autoLauncher.ts#L130

From your changes, appName now always return "subspace-desktop", from hardcoded or app.getName() 
https://github.com/subspace/subspace-desktop/blob/951577b367ba24a83e5d7e4dea324c8a3c0773ae/src/lib/autoLauncher.ts#L132

So instead of hardcoding we can just use this and will always get `subspace-desktop` from the tauri api.  

`this.appName = await app.getName()`

Maybe that can fix that missing **login_item** string.

ozgunozerk:
Yes Leo, I was running this under `dev`

PR #102, opened 2022-03-19
Launch on boot linux

1devNdogs:
# Summary

- Updated Linux methods to enable and disable autostart.

Tested .**deb** and .**AppImage** on `ubuntu 20.04,` the **autostart** works but the `--minimized` argument is not working.
Also checked #75  and it is working.


ozgunozerk:
Before these changes, it was working on macOS and Windows for this branch `launchOnBoot`. 
After these changes, hopefully, it should also work on Linux. 
Tonight, I will test this latest version in all 3 major platforms on my local machines. Will approve after that!

ozgunozerk (approved):

Issue #90, opened 2022-03-12
AppImage package on Ubuntu 21.10 doesn't sync


nazar-pc:
Nothing really happens, it just spins forever with this in the terminal:
```
(subspace-desktop:14549): Gdk-CRITICAL **: 03:17:09.547: gdk_window_thaw_toplevel_updates: assertion 'window->update_and_descendants_freeze_count > 0' failed
ATTENTION: default value of option mesa_glthread overridden by environment.
/usr/lib/x86_64-linux-gnu/gio/modules/libgiognutls.so: undefined symbol: g_tls_channel_binding_error_quark
Failed to load module: /usr/lib/x86_64-linux-gnu/gio/modules/libgiognutls.so
```

The last line is suspicious, but not necessarily the root cause.

<details>

![Знімок екрана з 2022-03-12 03-29-38](https://user-images.githubusercontent.com/928965/157998625-f1d55bbf-aac4-4c94-b2f4-a6ca35eeeb11.png)

</details>

ozgunozerk:
I tried just now. 
Started the application from `.AppImage` (downloaded from releases) on Ubuntu 20.04.3 LTS.
It synced without any problem, did not stuck anywhere. (deleted my `.subspace-desktop-farmer` folder prior to run)

Are you able to reproduce this again?

nazar-pc:
I'm on Ubuntu 21.10, latest stable release of the distro, not 20.04. I would still expect AppImage to be self-contained and work regardless.

nazar-pc:
OK, turns out it copies system libraries when building AppImage. So if package is built under Ubuntu 20.04, libraries are copied from 20.04. If we want to have packages for other versions we need to build it separately (Under Docker for instance).

1devNdogs:
If we fix  #97 the **AppImage** for **ubuntu 20.04 and 21.10** will be `correctly generated.` 
I found they already had this problem [here](https://github.com/tauri-apps/linuxdeploy-plugin-gtk/pull/1). 

- The tauri **beta** versions we use will run the `build` process using `linuxdeploy` and this script to load [GTK dependencies](https://github.com/linuxdeploy/linuxdeploy-plugin-gtk/blob/master/linuxdeploy-plugin-gtk.sh) to the AppImage.

- Using tauri **latests** versions will run the same `build` process, also using `linuxdeploy` **but using a custom script** to load
[...]

PR #103, opened 2022-03-20
Upgrade Tauri to latest version

nazar-pc:
I considered downgrading `zstd` version in a fork, but the simpler solution for now is to just disable compression.

Compression wasn't there before in Tauri (at least there was no such feature) and Debian package size got much smaller after upgrade anyway (30.2M -> 22.7M).

### DIFF
--- a/.env
+++ b/.env
@@ -1,4 +1,3 @@
 PUBLIC_API_WS="wss://aries-farm-rpc-b.subspace.network"
 LOCAL_API_WS="ws://localhost:9944"
-CONTEXT_MENU="OFF"
 DEFAULT_APP_DIR=".subspace-desktop-farmer"
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -65,7 +65,7 @@ jobs:
         run: yarn install --network-timeout 1000000 && yarn build
 
       - name: tauri run
-        uses: tauri-apps/tauri-action@v0.2
+        uses: tauri-apps/tauri-action@v0.3
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -27,10 +27,12 @@ jobs:
         with:
           node-version: 16
 
+      # TODO: when https://github.com/rust-lang/rust/issues/95267 and https://github.com/briansmith/ring/issues/1469
+      # are resolved, revert back to `nightly`
       - name: install Rust nightly
         uses: actions-rs/toolchain@v1
         with:
-          toolchain: nightly
+          toolchain: nightly-2022-02-15
           target: wasm32-unknown-unknown
           override: true
           components: rustfmt, clippy
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -65,7 +65,7 @@ jobs:
         run: yarn install --network-timeout 1000000 && yarn build
 
       - name: tauri run
-        uses: tauri-apps/tauri-action@v0.3
+        uses: tauri-apps/tauri-action@v0.2
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -56,7 +56,7 @@ jobs:
         run: yarn install --network-timeout 1000000 && yarn build
 
       - name: tauri run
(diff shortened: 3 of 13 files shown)


nazar-pc:
Please test if this actually works, I only checked that it compiles on Linux.

ozgunozerk (changes requested):
Unfortunately on macos, it creates the runtime error on quasar side:
`Scoped command osascript not found`. 
And produces an app with a blank screen with no functionality.
image

I will inspect this further, but this is not ready to merge yet.

nazar-pc:
For quasar we might need Leo's help

1devNdogs:
We need to update a few **settings** and **security checks** on **tauri.conf.json**. I have found at least these two as critical.

## TO FIX `fs scope` 

Solution:
- The `scope` error is related to [this](https://tauri.studio/docs/api/js/modules/fs#security) new configurations. Adding `scope` will be able to read and write files and dirs on it. On **Linux** at least I was getting similar errors and were solved by adding this  to **tauri > allowlist > fs > scope** config.

Two problems:
- The user will be **limited to this scope**, this will be a **problem** with selecting `non-HOME plot directory`, like another partition, external drives, etc. 
- Also, this configuration will be applied on the build and looks we can not update the scope after, still checking if there is another way to set this or maybe update the scope on user selection. I still can't find a solution to this.

@ozgunozerk  can you try adding this on tauri.conf.json 

`      
    "fs": {
        "all": true,
        "scope": ["$HOME/*"]
      }
`

## TO FIX `Shell scope:`

Looks we need to apply some confs also, but I can not test in here. @ozgunozerk this will probably solve the `applescript` execution problems. Just need to find the correct configuration.

https://tauri.studio/docs/api/js/modules/shell and https://tauri.studio/docs/api/js/modules/shell#security

1devNdogs:
Also check #90, will be fixed automatically if we get this upgrade running.

ozgunozerk:
tauri action v0.3 creates problem. I believe it is not yet stable (also it is not `latest` in the release section of their repo). Switching back to v0.2 for now.

ozgunozerk:
with march 03-23 update on rust nightly, `ring` create is creating problems on the building phase.
<details>
error: no rules expected the token `aarch64_apple`
   --> /Users/ozgunozerk/.cargo/registry/src/github.com-1ecc6299db9ec823/ring-0.16.20/src/cpu.rs:257:13
    |
165 |     macro_rules! features {
    |     --------------------- when calling this macro
...
257 |             aarch64_apple: true,
    |             ^^^^^^^^^^^^^ no rules expected this token in macro call

error[E0425]: cannot find value `AES` in module `cpu::arm`
   --> /Users/ozgunozerk/.cargo/registry/src/github.com-1ecc6299db9ec823/ring-0.16.20/src/aead/aes.rs:381:65
    |
381 |         if cpu::intel::AES.available(cpu_features) || cpu::arm::AES.available(cpu_features) {
    |                                                                 ^^^ not found in `cpu::arm`
    |
help: consider importing this constant
    |
15  | use crate::cpu::intel::AES;
    |

error[E0425]: cannot find value `PMULL` in module `cpu::arm`
   --> /Users/ozgunozerk/.cargo/registry/src/github.com-1ecc6299db9ec823/ring-0.16.20/src/aead/gcm.rs:315:26
    |
315 |             || cpu::arm::PMULL.available(cpu_features)
    |                          ^^^^^ not found in `cpu::arm`

error[E0425]: cannot find value `ARMCAP_STATIC` in this scope
   --> /Users/ozgunozerk/.cargo/registry/src/github.com-1ecc6299db9ec823/ring-0.16.20/src/cpu.rs:235:41
    |
235 |             if self.mask == self.mask & ARMCAP_STATIC {
    |                                         ^^^^^^^^^^^^^ not found in this scope

error[E0425]: cannot find value `ARMCAP_STATIC` in this scope
   --> /Users/ozgunozerk/.cargo/registry/src/github.com-1ecc6299db9ec823/ring-0.16.20/src/cpu.rs:288:36
    |
288 |     static mut GFp_armcap_P: u32 = ARMCAP_STATIC;
    |                                    ^^^^^^^^^^^^^ not found in this scope
</details>

nazar-pc:
Here is a report in:
* ring: https://github.com/briansmith/ring/issues/1469
* rust: https://github.com/rust-lang/rust/issues/95267

ozgunozerk:
Thanks Nazar, I've opened a issue about `ring`, I will keep an eye on these threads you provided and update to `nightly` when this issue is resolved. For now, I believe it is ok to continue with 2022-02-15 version of rust.

ozgunozerk:
Tested on macos and windows, it is working

ozgunozerk (approved):

Issue #88, opened 2022-03-12
Creates default directory without permission


nazar-pc:
When app starts it creates `/home/nazar-pc/.subspace-desktop-farmer` even if I didn't ask it to and didn't confirm that this is where I want to store data. This is unexpected and should be fixed.

Issue #87, opened 2022-03-12
Can't specify folder that doesn't exist


nazar-pc:
When starting an app, default directory is `/home/nazar-pc/.subspace-desktop-farmer`. It, obviously, doesn't exist initially.

However as soon as I change it to anything, let's say `/home/nazar-pc/.subspace-desktop-farmer2` it says "Invalid Directory".

This doesn't make sense and is a bug. It is can create a directory that is one level deep from existing by default, it should be capable of creating custom directories as well.

ozgunozerk:
Yes, the behaviour is conflicting I agree. It can create the directory for the default path, but for any other folder, it expects that folder to be there. We can fix this easily, thanks for pointing that out. I had forgotten about it.

nazar-pc:
No directories need to be created until confirmed by user IMHO, otherwise garbage empty directory will be left, see #88 and #85

Issue #85, opened 2022-03-12
Default path is wrong


nazar-pc:
Default path on Linux is suggested to be this:
> /home/nazar-pc/.subspace-desktop-farmer

This is wrong and should be fixed. Configs (if any) should go to `~/.config/subspace-desktop` and data to `~./local/share/subspace-desktop` following XDG specification. I can imagine our desktop app pollutes home folders on other platforms in a similar way.

PR #108, opened 2022-03-23
Update app directories

1devNdogs:
# Summary.

- #85
  - Use OS `configDir` to store config file 
  - Use OS `dataDir` as the default plot directory. 
- #88 
  - On app launch no app default data directories are created. 
- #87
  - On `SetupPlot` page, a confirmation from the user will be required to **create** any **custom** or **nonexisting** directory.  

# Details.
### src/lib/
- **util** - move methods related to own file:
- /**appConfig** - Use LocalStorage to persist AppConfig.
- /**appData** - Handle plot and node data directory, will use OS `dataDir` by default or a custom selected directory.
- /**appData - dialogs** - Util to display, `existingDirectoryConfirm`, `newDirectoryConfirm`, `emptyDirectoryInfo` 

### src/pages/
 - Pages are updated to work using new directories.

## Test
Need help to test this. 

- [x]  Ubuntu
- [x]  Windows
- [x]  MacOS


ozgunozerk:
On macos -> 
`username/Library/Application Support/subspace-desktop/`
this folder is created automatically when the app starts without any confirmation. Furthermore, app doesn't want to continue after **SetupPlot** page because this directory is not empty (there is `config.json` file there at this point). 
image

I've tried deleting the aforementioned folder under `Application Support` and restarting the app. The issue persists (it is automatically created by the program).

1devNdogs:
Looks like for `macOS` the Tauri API returns the same OS directory for `configs` and `application data. 
For Windows and Linux this should not happen. 

- `tauri.path.dataDir` macOS:** Resolves to `$HOME/Library/Application Support
- `tauri.path.configDir`:  macOS: Resolves to $HOME/Library/Application Support

let me check **macOS library docs** for the correct use of `Application Support` directory.

1devNdogs:
Will send an update following developer docs.



https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/MacOSXDirectories/MacOSXDirectories.html
> Application Support 

> Contains all app-specific data and support files. These are the files that your app creates and manages on behalf of the user and can include files that contain user data.

> By convention, all of these items should be put in a subdirectory whose name matches the bundle identifier of the app. For example, if your app is named MyApp and has the bundle identifier com.example.MyApp, you would put your app’s user-specific data files and resources in the ~/Library/Application Support/com.example.MyApp/ directory. Your app is responsible for creating this directory as needed.

src/lib/autoLauncher.ts
```diff
@@ -63,6 +63,14 @@ Icon=PATH_TO_APP_ICON
     const autostartDirectory = (await path.configDir()) + "/autostart"
     const autostartAppFile = autostartDirectory + "/" + appName + ".desktop"
     return autostartAppFile
+  },
+  async createAutostartDir(appName: string): Promise<string> {
+    const autostartDirectory = (await path.configDir()) + "autostart"
```
nazar-pc, line 68: Above function has `/` on this line and this one doesn't. They shouldn't be both correct at the same time.

src/lib/directories/configFile.ts
```diff
@@ -0,0 +1,56 @@
...
+    for (const [key, value] of Object.entries(newData)) {
+      Object.assign(configFile[key], value)
+    }
+    await this.writeConfigFile(configFile)
+  },
+  async writeConfigFile(config: ConfigFile): Promise<void> {
+    const configDir = (await tauri.path.configDir()) + dirName
+    await tauri.fs
+      .writeFile({
+        path: configDir + "/config.json",
```
nazar-pc, line 37: Are you sure you need a config file? Local Storage seems to be persistent, maybe use that instead? Or these are for fundamentally different purpose? (I already saw some of the similar data stored in Local Storage and config file (like path to config is in Local Storage).

src/lib/autoLauncher.ts
1devNdogs: I am currently working on merge `tauri-upgrade` to this branch. Will be fixed on the next push.  ThanksI

src/lib/directories/configFile.ts
1devNdogs: > path to config is in Local Storage.

- Should not be in this branch. It is not required anymore as we now use `OS config directory`, the app always knows where the **config.json file lives**.

> Are you sure you need a config file? Local Storage seems to be persistent, maybe use that instead?

- 👍 agree with replacing the use of `config.json` and using **LocalStorage** instead.
Will remove a **lot of code**, **few files**, will **reduce code** on pages working on this file.
And finally as an extra, will **fix the macOS problem** on using the same OS directory for **config** and default application **data**.

> This config.json came from the initial code, the demo version John develop. I initially assumed that it was necessary for **tauri or quasar**, but after working in the app, removing data from it, updating the structure also I have realized that NOW is more a mini **util/cache** than a **config** itself.

- Do we move it to LocalStorage? I say yes and with a big smile: One less folder to handle, less invasive to the filesystem, and actually, will fix this problem on macOS using same directory for config and app data.

ozgunozerk:
Works on macOS (at least on my machine).
However, after the config change, do we need to re-check linux and windows?

ozgunozerk (approved):

PR #111, opened 2022-03-24
Fix for Linux autostart icon

nazar-pc:
Should work fine for Debian package, will not work for AppImage, but it didn't work before, so fine.

BTW, I noticed it actually doesn't create the autostart file on Linux for some reason (on 0.2.0).

### DIFF
--- a/src/lib/autoLauncher.ts
+++ b/src/lib/autoLauncher.ts
@@ -29,14 +29,13 @@ const linAL = {
     const autostartAppFile = await this.getAutostartFilePath(appName)
     const response: ChildReturnData = { stderr: [], stdout: [] }
     const hiddenArg = minimized ? ' --minimized' : '';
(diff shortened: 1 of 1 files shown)


1devNdogs:
> BTW, I noticed it actually doesn't create the autostart file on Linux for some reason (on 0.2.0).

We got that covered, will be fixed after https://github.com/subspace/subspace-desktop/pull/108. Still testing to fix `macOS` directories, I'am also double checking on a windows machine. 

Will appreciate if you can have a fast look on [#108](https://github.com/subspace/subspace-desktop/pull/108#issuecomment-1077054581), about the way to handle **macOS** default directories.

1devNdogs:
I updated with latest changes from main, after that,  generated a clean **build**. 
I tested `.deb` and `.AppImage` on **20.04**. Both bundles got autostart working, tested several times and looks fine. 
 
image

1devNdogs (approved):

PR #114, opened 2022-03-25
disable advanced page, and small tweaks

ozgunozerk:
This PR is doing mainly 3 things:
- removes advanced page, and hence mnemonic import option
- turn `tSSC` instances into `testnetSSC`
- readme update (small description)

Readme needs more improvement, we will work on that with @ImmaZoni later.


src-tauri/chain-spec.json
```diff
@@ -17,7 +17,7 @@
   "properties": {
     "ss58Format": 2254,
     "tokenDecimals": 18,
-    "tokenSymbol": "tSSC"
+    "tokenSymbol": "testnetSSC"
```
nazar-pc, line 20: Why this random change? It is called tSSC everywhere else (including Polkadot.js apps explorer and ss58-registry crate).

ozgunozerk: This is what we discussed for the last 2 meetings. Jeremiah wanted us to be VERY explicit about these tokens have no value. 
But the symbol should have stayed same in this place, it's a victim of `replaceAll`. Reverting that specific instance

nazar-pc (approved):

Issue #128, opened 2022-03-31
Estimating Plot Size takes too much time


ozgunozerk:
We have to somehow enhance this. It is bottlenecking the development process as well

ozgunozerk:
If we are syncing from bootstrap, change it to sync from peers

ImmaZoni:
The root cause of this should be our main focus. (imo)

This seems to be correlated to the domain/bootstrap node.

For example. 
Earlier I could not load the Explorer on my browser (30min+)
At this same time I was unable to run subspace-desktop through setup because it was hanging on querying the estimated size.

Shouldn't this be queried with the local node instead of relying on the explorer?

1devNdogs:
Now it will take no time.

ozgunozerk:
Still takes too much time. Tried on MacOS, builded with `yarn build`, it was `main` branch

PR #131, opened 2022-04-02
Fix slow plot size estimation.

1devNdogs:
# Sumary

- Fixes: #128 

It now gets the value from the `subspace storage`.
Should take less than a second now.

### DIFF
--- a/src/lib/client.ts
+++ b/src/lib/client.ts
@@ -1,7 +1,7 @@
 import { ApiPromise, WsProvider } from "@polkadot/api"
 import type { Vec } from "@polkadot/types/codec"
-import type { u128 } from "@polkadot/types"
-import type { Hash, AccountId32 } from "@polkadot/types/interfaces"
+import type { u128, u32 } from "@polkadot/types"
+import type { AccountId32 } from "@polkadot/types/interfaces"
 import * as event from "@tauri-apps/api/event"
 import { invoke } from "@tauri-apps/api/tauri"
 import { reactive } from "vue"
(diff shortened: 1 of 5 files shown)


nazar-pc (approved):
New code makes perfect sense.
Old code is just :rofl:

src/lib/client.ts
```diff
@@ -190,28 +190,9 @@ export class Client {
...
-          return segmentIndex <= 1 ? 1 : segmentIndex
-        }
-      }
-      return await this.getNetworkSegmentIndex(
-        signedBlock.block.header.parentHash
-      )
-    }
+  public async getNetworkSegmentIndex(): Promise<number> {
+    const recordsRoot = await this.publicApi.query.subspace.counterForRecordsRoot() as u32
+    return recordsRoot.toNumber()
```
nazar-pc, line 195: This is not index though, this is number of elements

nazar-pc (changes requested):

src/lib/client.ts
1devNdogs: Sure, I had to update several references but at the same time, fixed a few weird variable names. looks better this way.

src/lib/client.ts
```diff
@@ -183,14 +183,12 @@ export class Client {
     return signedBlock.block.header.number.toNumber()
   }
 
-  /* SEGMENT INDEX */
-  // Used to check archived segments locally and display farmer plot status.
-  public async getLocalFarmerSegmentIndex(): Promise<number> {
+  public async getLocalSegmentCount(): Promise<number> {
     const plot_progress_tracker =
       ((await invoke("plot_progress_tracker")) as number) / 256
```
nazar-pc, line 188: This will work, but there are a few issues:
* `plot_progress_tracker` is a bad name, should be `get_plotted_pieces_count()` or something like that
* `256` will work for now, but it is not guaranteed to be that way in the future, the actual value must be extracted from the runtime: https://github.com/subspace/subspace/blob/main/crates/sc-consensus-subspace/src/lib.rs#L913
* as the result of above and considering that you should already be connected to local node anyway I suggest the following:
    * subscribe to events of local node and monitor for root block archiving that contains archived segment in it, making UI reactive rather than relying on timers
    * remove `plot_progress_tracker` on backend

Non-blocking, can be done in a separate PR, just don't forget about this.

cc @ozgunozerk

nazar-pc (approved):

src/lib/client.ts
nazar-pc: Thought some more about this. If we care about plotting as such, we do have to inspect the plot and not the chain, but I'm not sure there is a major difference between those right now until we implement DSN sync of the farmer.

ozgunozerk: I've opened the issue #133 for tracking this

Issue #74, opened 2022-03-10
Redirect the user to block explorer from rewards in the dashboard


ozgunozerk:
Make some clickable links in the dashboard, so that user can view his public key and rewards in the block explorer

PR #132, opened 2022-04-02
Reward address updates

1devNdogs:
# Summary

@ozgunozerk I have updated your branch with the **latest** changes from main **main**. Added some updates to improve UI.
Created a PR to main, less changes to compare

## Updates.
- #74 
    - Add a `link` to Polkadot explorer on the target block,  I think it was necessary to cover this issue here, as it helps to **validate** the block and the **farmer and reward addresses**.

- To make data fit on the table, updated the `time` column using a date format and made a few adjustments to it.


ozgunozerk:
We have decided not to confuse the end-user with the `signingKey`, and only expose `rewardAddress`. And even though it is not 100% accurate, we will display the `rewardAddress` under `farmedBy` instead of having a new column. So it will be simpler and provide better UX. 
I'm going to change it now.
Thanks for all the wonderful changes though, helps a lot Leo!

ozgunozerk (dismissed):

ozgunozerk (approved):

Issue #127, opened 2022-03-31
keyboard shortcuts does not work


ozgunozerk:
`ctrl+c` or `crtl+v` and etc. are not working in the app (at least in dev mode)

ozgunozerk:
May be related to #96 as well

1devNdogs:
Can you have a look at this? https://github.com/tauri-apps/tauri/issues/2397

For me, windows and linux got shortcuts working.

ozgunozerk:
Thanks Leo, that looks promising, adding menu seems to be the right solutions, since the topbar menu is not working in mac as well. I think it will solve many problems at once for mac

PR #135, opened 2022-04-02
Add menu for mac

### LINKED ISSUE #96: `cmd+q` keyboard shortcut for quitting the app does not work (macos)

ozgunozerk:
It is working for macos. 
I don't know how it looks on linux and windows, will try soon.

Fixes #96, #127 

Inspect all changes at once, since it is small and I will squash merge anyways


src-tauri/src/menu.rs
```diff
@@ -0,0 +1,48 @@
...
+            .add_native_item(MenuItem::Redo)
+            .add_native_item(MenuItem::Separator)
+            .add_native_item(MenuItem::Cut)
+            .add_native_item(MenuItem::Copy)
+            .add_native_item(MenuItem::Paste)
+            .add_native_item(MenuItem::SelectAll);
+
+        let window_menu = Menu::new()
+            .add_native_item(MenuItem::Minimize)
+            .add_native_item(MenuItem::CloseWindow);
```
nazar-pc, line 36: Why do we need `edit` menu? Also it seems that `my_app` and `window` are very similar if not the same.

ozgunozerk: We need edit menu, since if we do not add these native items, we cannot use the shortcuts on mac. Unfortunately that's how it works. And if they all populate the same submenu, it looks really really bad. 
So the convention for macos applications is storing these (cut, copy, etc.) under edit menu
image
this is also the case in the other apps I've used like discord etc. I just followed the convention for being more user-friendly

nazar-pc: Hm, edit menu makes sense, though this is definitely a strange macOS design decision. What about App vs Window?

ozgunozerk: It is actually the same reason, following the convention. I agree with you they are being similar `App` and `Window`. But for **minimize, zoom, etc.** these are stored in `Window` submenu. On the other hand, **hide and quit** are stored in the main submenu, which is the app itself (**subspace-desktop** in our case). 
image
image

nazar-pc (approved):

PR #139, opened 2022-04-03
autoLaunch fix and version bump

ozgunozerk:
AutoLaunch bug (not remembering the config) was fixed before. But after local storage changes, it regressed and reappeared. 

The fix was pretty simple (inspect appConfig.ts file). Now it should be fine, and more safe (there are few improvements on `enable` and `disable` functions.

Other than that, the naming was confusing and not making sense for `autoLaunch` related variables in `mainMenu`. 

Lastly, bumped the version to 0.3.0 after massive changes we had recently, and will publish a pre-release after this PR

### DIFF
--- a/src/lib/appConfig.ts
+++ b/src/lib/appConfig.ts
@@ -22,7 +22,7 @@ export const appConfig = {
       if (plot) newAppConfig.plot = plot
       if (account) newAppConfig.account = account
       if (segmentCache) newAppConfig.segmentCache = segmentCache
-      if (launchOnBoot) newAppConfig.launchOnBoot = launchOnBoot
+      if (launchOnBoot != null) newAppConfig.launchOnBoot = launchOnBoot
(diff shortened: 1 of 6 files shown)


src/lib/appConfig.ts
```diff
@@ -22,7 +22,7 @@ export const appConfig = {
       if (plot) newAppConfig.plot = plot
       if (account) newAppConfig.account = account
       if (segmentCache) newAppConfig.segmentCache = segmentCache
-      if (launchOnBoot) newAppConfig.launchOnBoot = launchOnBoot
+      if (launchOnBoot != null) newAppConfig.launchOnBoot = launchOnBoot
```
ozgunozerk, line 25: Here is the fix. 
Since `launchOnBoot` is **boolean**, when provided the value `false`, it was not updating the **config**.... 👻

nazar-pc (approved):

PR #140, opened 2022-04-03
notify for autolaunch

### LINKED ISSUE #136: Autolaunch added without any warning
I just started subspace-desktop 0.2.2 and it added itself to autostart. I disabled it, closed without going through setup process and started again, it added itself to autostart again :angry: 

Not only it should ask the user, it should also remember the decision and not add new entries again and again on every start.

* [ ] Do not enable autostart without permission
* [x] Remember decision and don't add autostart entry again and again

Maybe some of this was already fixed in `main`, I didn't check.

ozgunozerk:
Waiting for #139 to merge (for remembering the LaunchOnBoot preference bug fix)

Fixes #136 .

Did not want to complicate the application by prompting the user for permission on this. Instead, we are notifying the user about the app will be launched on boot, and if wanted, it can be disabled from settings.

### DIFF
--- a/src/pages/Index.vue
+++ b/src/pages/Index.vue
@@ -25,7 +25,7 @@ import { defineComponent } from "vue"
 import { globalState as global } from "src/lib/global"
 import * as util from "src/lib/util"
 import { appConfig } from "src/lib/appConfig"
-import { LocalStorage } from "quasar"
+import { Notify } from "quasar"
 const lang = global.data.loc.text.index
 
 export default defineComponent({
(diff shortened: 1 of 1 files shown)


ozgunozerk:
image
this triggers only on the `Welcome` page and only if the `LaunchOnBoot` is set to true (and it is set to true automatically). 
The important part is:
*if the user sets `LaunchOnBoot` to `false`, and closes the application (without even starting to plot), and re-opens the application, the app will start from the `Welcome` page, but since `LaunchOnBoot` is disabled previously, this notification will not be shown.

nazar-pc:
I thought it would be better to not start it automatically by default in the first place :thinking:

ozgunozerk:
In the desktop app meetings, Jeremiah suggested auto launch should be enabled by default. Thinking that this will be farming app, users probably will want to keep it running whilst the PC is on. I think its a good idea to enable this option by default

nazar-pc (approved):

PR #147, opened 2022-04-07
chainspec update

ozgunozerk:


src-tauri/chain-spec.json
```diff
@@ -11,6 +11,10 @@
     [
       "/dns/telemetry.polkadot.io/tcp/443/x-parity-wss/%2Fsubmit%2F",
       0
+    ],
+    [
+      "/dns/telemetry.subspace.network/tcp/443/x-parity-wss/%2Fsubmit%2F",
+      0
```
ozgunozerk, line 17: should this be 1?

nazar-pc: 0 is sufficient, to show the node, but I didn't check what extra data are sent with 1. Probably fine as is.

nazar-pc (approved):

PR #152, opened 2022-04-09
Fix linux tray icon

ozgunozerk:
bumps version to 0.3.3

fixes the linux bug for icon (tested on my ubuntu, it was working)

Thanks to @1devNdogs for fixing it!


nazar-pc (approved):
What did you fix though? Looks like image was reverted to the previous one, but it is the one that is hard to see now. Where is the problem?

ozgunozerk:
Quoting from what Leo explained yesterday to me:
> 32x32 icon is not correctly generated from tauricon as this new black stroke on subspace-logo.png is tooo tiny and that on a low size image produce a corrupted png, the good thing is that this 32x32.png is only used on ubuntu. For system tray icon.
> And system tray is always dark on Ubuntu so we can keep the older 32x32.png and new bordered icon everywhere

nazar-pc:
> 32x32 icon is not correctly generated from tauricon as this new black stroke on subspace-logo.png is tooo tiny and that on a low size image produce a corrupted png, the good thing is that this 32x32.png is only used on ubuntu. For system tray icon.

This doesn't make sense to me. Why can't we generate non-corrupted PNG?

ozgunozerk:
Here is the guide we have followed always: https://tauri.studio/docs/guides/icons

We have tried many versions of the logo that Diana sent us. Unfortunately, we were not able to make it work under ubuntu. I tried many times, and Leo spent the night I believe on this issue. 
The solution he could came up with was this. 
Although there can be a better solution and there are things that we do not fully understand, this IS a solution in the end imho.

nazar-pc (approved):

PR #159, opened 2022-04-11
reward address is displayed

### LINKED ISSUE #157: I don't see where the address is displayed
I don't see where the address is displayed, only have seed phrase, can you help me find my address by subspace-desktop

ozgunozerk:
Trivial PR for frontend.
Displays the reward address in the dashboard statically (even when there are 0 block farmed)

Fixes #157 
Screen Shot 2022-04-11 at 13 58 25

### DIFF
--- a/src/components/farmedList.vue
+++ b/src/components/farmedList.vue
@@ -6,11 +6,14 @@ q-card(bordered flat)
         .row.items-center
           q-icon.q-mr-sm(color="grey" name="grid_view" size="40px")
           .text-h6.text-weight-light {{ lang.farmedBlocks }}:
-      .col-auto.q-mr-md
+      .col-auto.q-mr-xl
         h6 {{ farmedBlocksList?.length }}
-      .col-auto.q-mr-md
+      .col-auto.q-mr-xl
         .text-weight-light Total Earned
         p {{ farmedTotalEarned }} testnetSSC
+      .col-auto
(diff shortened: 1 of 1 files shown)


src/components/farmedList.vue
```diff
@@ -87,7 +95,28 @@ export default defineComponent({
...
+      if (addr == null) {
+        const config = appConfig.getAppConfig()
+        if (config) {
+          const { farmerPublicKey } = config.account
+          const addr2: string = addr ?? farmerPublicKey
+          return this.rewardAddress = addr2
+        }
+      return this.rewardAddress = ""
+      } else {
+        return this.rewardAddress = addr
```
nazar-pc, line 110: 1) Indentation is odd
2) Don't mix assignment and return, please

nazar-pc (commented):
On a general node, try to use present tense of what needs to be done in commit messages instead, for instance:
> display reward address

src/components/farmedList.vue
ozgunozerk: It wasn't working when I deleted the `return`. Will try to solve it, else I will ask help from you or Serge.

nazar-pc: Well, you can always split those into 2 lines:
```typescript
this.rewardAddress = addr
return this.rewardAddress
```

nazar-pc (approved):

PR #160, opened 2022-04-11
Drop public rpc connection

ozgunozerk:
This PR will make the desktop app disconnect from the PublicRPC once the connection is not needed.

Please refer to the comments below for additional details

### DIFF
--- a/src/lib/client.ts
+++ b/src/lib/client.ts
@@ -169,6 +169,13 @@ export class Client {
     await this.publicApi.isReady
   }
 
+   /* Disconnects from PUBLIC-rpc node - Example: farm-rpc.subspace.network */
+  public async disconnectPublicApi(): Promise<void> {
+    if (this.publicApi.isConnected) {
+      await this.publicApi.disconnect()
+    }
+  }
+
   public async getBlocksData(): Promise<[number, number]> {
     const blocksNumbers = await Promise.all([
       this.getLocalLastBlockNumber(),
@@ -207,7 +214,7 @@ export class Client {
     await this.connectLocalApi()
     return clientIdentity
   }
-  
+
   // TODO: Disable mnemonic return from tauri commmand instead of this validation.
   private async startNode(path: string): Promise<ClientIdentity> {
     const { publicKey, mnemonic } = await tauri.invoke("start_node", { path })
--- a/src/lib/client.ts
+++ b/src/lib/client.ts
@@ -218,7 +218,7 @@ export class Client {
   // TODO: Disable mnemonic return from tauri commmand instead of this validation.
   private async startNode(path: string): Promise<ClientIdentity> {
     const { publicKey, mnemonic } = await tauri.invoke("start_node", { path })
-    const farmerPublicKey: AccountId32 = this.publicApi.registry.createType(
+    const farmerPublicKey: AccountId32 = this.localApi.registry.createType(
       "AccountId32",
       publicKey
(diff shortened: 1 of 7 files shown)


src/lib/client.ts
```diff
@@ -169,6 +169,13 @@ export class Client {
     await this.publicApi.isReady
   }
 
+   /* Disconnects from PUBLIC-rpc node - Example: farm-rpc.subspace.network */
+  public async disconnectPublicApi(): Promise<void> {
+    if (this.publicApi.isConnected) {
+      await this.publicApi.disconnect()
+    }
```
nazar-pc, line 176: I'm not sure check for connected is necessary. It may be racy with this check in place.

ozgunozerk:
PublicRPC connection is used in querying the blockchain size. After retrieving the data from publicRPC, the connection is dropped. ✅

ozgunozerk:
PublicRPC is used also in here when starting the node: https://github.com/subspace/subspace-desktop/blob/drop-publicRPC-connection/src/lib/client.ts#L219

I did not understand the purpose of this. @1devNdogs can you explain why we need `const farmerPublicKey: AccountId32 = this.publicApi.registry.createType(
      "AccountId32",
      publicKey
    )` ?

src/pages/Dashboard.vue
```diff
@@ -97,6 +97,7 @@ export default defineComponent({
       this.global.status.message = lang.syncedMsg
       this.checkNodeAndNetwork()
       this.checkFarmerAndPlot()
+      this.client.disconnectPublicApi()
```
nazar-pc, line 100: Why does this component need to connect to public API at all?

nazar-pc (commented):
It looks like public API should be just a local variable in those contexts where it si needed instead of some global entity, isn't it?

ozgunozerk:
PublicRPC is also used for querying the lastBlockNumber, so it can compare this with the localNode's last block number. And notify the user about the sync status (whether completed or in progress). 

And the last place PublicRPC is used is: getting the blockchainSize again for determining the plot size (this is eliminated with the partial plotting PR).

However, both of these are done sequentially, and the connection is now drop after both queries ✅

src/pages/Dashboard.vue
ozgunozerk: See the below comment. 
`checkNodeandNetwork` checks for last block from the publicRPC to inform the user on sync status.
`checkFarmerAndPlot` retrieves the blockchain size from the publicRPC, but it will be gone after partial replica PR (we will be querying the plot size from config, because user gets to choose it)

src/lib/client.ts
ozgunozerk: ok, i will change it, no problem :)

src/pages/Dashboard.vue
nazar-pc: Does this mean public RPC connection can be established and closed in 2 places, but it is the same global variable? Looks like a recipe for confusing bugs.

ozgunozerk:
> It looks like public API should be just a local variable in those contexts where it is needed instead of some global entity, isn't it?

Leo is already working on a big refactoring of the front end. He may be fixing that structure as you suggested :)

nazar-pc:
Since memory was leaking on relayer as well, I suspect this might fix memory leak we see in desktop app as well

nazar-pc (approved):

PR #162, opened 2022-04-13
mnemonic workflow fixed

### LINKED ISSUE #163: User may skip the mnemonic and will not be able to display it again
## Problem Description
1. user starts the app for the first time
2. proceeds to plotting screen
3. in plotting screen, the user does not display the intro (hence, mnemonic is not shown)
4. user quits the app for god knows why
5. when user re-opens the app, it goes directly to the dashboard, because the config is already created
6. there are no functionality to display mnemonic in the dashboard, and in order to have that, we have to store mnemonic

## Potential solution
- we can move the mnemonic modal to a page prior to plotting
- we can store the mnemonic in keychain

### LINKED ISSUE #164: Hints on the plotting page can only be displayed once
User should be able to show hints as many times as s/he likes

ozgunozerk:
This PR fixes an important flaw in the workflow:
**user may not have displayed mnemonic, and lose the chance of displaying mnemonic ever again**
fixes #164 
fixes #163 

Nazar suggested that we can store the mnemonic in the keychain, but I'm still favouring not storing the mnemonic at all. 
*And even if we decide to store the mnemonic in the keychain, it will be implemented later. And this problem needs to be addressed ASAP imo*. So, in short, this will either be a temporary solution, or a permanent one.

### Problem Description
1. user starts the app for the first time
2. proceeds to plotting screen
3. in plotting screen, the user does not display the intro (hence, mnemonic is not shown)
4. user quits the app for god knows why
5. when user re-opens the app, it goes directly to the dashboard, because the config is already created
6. there are no functionality to display mnemonic in the dashboard, and in order to have that, we have to store mnemonic

### Mindset behind this solution
In all the crypto-related apps I've seen that includes creating mnemonic:
**They are not starting the functionality before giving the mnemonic to the user**.
So, I decided to follow the same path. Imo, we should not start plotting/farming unless the user confirms on the mnemonic.

### The Solution
Moved the mnemonic confirmation modal to the end of the `SetupPlot` page.
Now, immediately after confirming the directory, a new modal pops up about mnemonic. And unless the user confirms that s/he copied the mnemonic, it is impossible to proceed into the `plotting` page.

If the user confirms the directory, but then quits the app without confirming the mnemonic modal, it does not matter :)
The app will start from scratch again, because neither node nor farmer had started.

**this new mnemonic modal is skipped totally if user imports reward address**

### What else changed?
To be able to get the mnemonic, we have to create the identity. So there is a new function in the backend just for creating an identity (we don't want to start node or farmer to just get the identity). 
`startNode` is not returning mnemonic anymore (it was a hacky solution anyway), it is the new `create_identity` functions responsibility.

The modal in the plotting screen is separated from the `next` button. Now its only purpose is to show hints. So there is new `hints` button in the plotting screen, which can be clicked unlimitedly. `next` button is disabled, and will be enabled only when plotting is finished.

In summary, we have a less complex and more direct workflow now.

### DIFF
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -10,7 +10,6 @@ mod menu;
 mod node;
 
 use anyhow::Result;
-use bip39::{Language, Mnemonic};
 use log::{debug, error, info};
 use serde::Serialize;
 use std::path::PathBuf;
@@ -69,11 +68,6 @@ async fn start_node(path: String) -> [u8; 32] {
 }
 
 #[tauri::command]
-fn identity_create(path: String) -> String {
-    create_identity(path.into()).unwrap()
-}
-
-#[tauri::command]
 fn get_disk_stats(dir: String) -> DiskStats {
     debug!("{}", dir);
     let free: u64 = fs2::available_space(&dir).expect("error");
@@ -129,8 +123,7 @@ async fn main() -> Result<()> {
                 get_this_binary,
                 farming,
                 plot_progress_tracker,
-                start_node,
-                identity_create
+                start_node
             ],
             #[cfg(target_os = "windows")]
             tauri::generate_handler![
@@ -141,8 +134,7 @@ async fn main() -> Result<()> {
                 get_disk_stats,
                 farming,
                 plot_progress_tracker,
-                start_node,
-                identity_create
+                start_node
             ],
         )
         .build(tauri::generate_context!())
@@ -197,16 +189,6 @@ async fn init_node(base_directory: PathBuf) -> Result<[u8; 32]> {
     });
 
     Ok(public_key)
-}
-
-fn create_identity(base_directory: PathBuf) -> Result<String> {
-    let identity = Identity::open_or_create(&base_directory)?;
-
-    Ok(
-        Mnemonic::from_entropy(identity.entropy(), Language::English)
-            .unwrap()
-            .into_phrase(),
-    )
 }
 
 /// Start farming by using plot in specified path and connecting to WebSocket server at specified
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -33,13 +33,6 @@ static PLOTTED_PIECES: AtomicUsize = AtomicUsize::new(0);
 const BEST_BLOCK_NUMBER_CHECK_INTERVAL: Duration = Duration::from_secs(5);
 
 #[derive(Serialize)]
-#[serde(rename_all = "camelCase")]
-struct FarmerIdentity {
-    public_key: [u8; 32],
-    mnemonic: String,
-}
-
-#[derive(Serialize)]
 struct DiskStats {
     free_bytes: u64,
     total_bytes: u64,
@@ -71,8 +64,13 @@ async fn farming(path: String, reward_address: String, plot_size: u64) {
 }
 
 #[tauri::command]
-async fn start_node(path: String) -> FarmerIdentity {
+async fn start_node(path: String) -> [u8; 32] {
     init_node(path.into()).await.unwrap()
+}
+
+#[tauri::command]
+fn identity_create(path: String) -> String {
+    create_identity(path.into()).unwrap()
 }
 
 #[tauri::command]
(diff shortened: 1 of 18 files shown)


nazar-pc (commented):
Do we need to generate identity on the backend at all? Maybe we can generate it on the frontend with Polkadot.js and just send reward address to the backend when farmer needs to start? This will unify logic between user-provided and auto-generated reward address.

Identity for the plot should be auto-generated and never exposed to the frontend I think.

ozgunozerk:
I also thought about that, and unifying the logic sounded very pleasing to me as well. 
However, what will be the advantage of generating it from polkadot on the frontend? 
If we prefer generating an identity on the frontend, we will still have to display the mnemonic to the user, and then pass the reward address to the farming, and for the users who did not select `import reward address` option, there will be confusion on the block explorer (the farming address and the reward address will be different, although they did not do anything custom). I think this would seem weird.

Hence, I went with generating identity on the backend. It is more or less the same process (mnemonic, private key and public key generation), and on top of that, the farmer and reward address will be the same for the ones who went with `Quick Start`. And the last benefit is for me: I know how to generate an identity with our backend, but have to do some research and googling for generating polkadot address :)

nazar-pc:
>  However, what will be the advantage of generating it from polkadot on the frontend?

Right now there are 2 code paths on the backend and 2 on the frontend, with keypair generation on the frontend backend will only have one path, while frontend will be easier due to no need to make another "ffi" call into the backend just to get a keypair.

> there will be confusion on the block explorer (the farming address and the reward address will be different, although they did not do anything custom). I think this would seem weird.

This is not weird and largely irrelevant since with multi-plotting every plot will have unique auto-generated identity anyway.

src-tauri/src/main.rs
```diff
@@ -10,7 +10,6 @@ mod menu;
 mod node;
 
 use anyhow::Result;
-use bip39::{Language, Mnemonic};
```
nazar-pc, line 13: Can we delete it from dependencies then?

src-tauri/src/main.rs
```diff
@@ -71,7 +63,7 @@ async fn farming(path: String, reward_address: String, plot_size: u64) {
 }
 
 #[tauri::command]
-async fn start_node(path: String) -> FarmerIdentity {
+async fn start_node(path: String) -> [u8; 32] {
```
nazar-pc, line 66: Does it still need to return anything with recent changes?

src/lib/client.ts
```diff
@@ -228,6 +228,16 @@ export class Client {
     this.data.farming.farmed = this.farmed
   }
 
+  public async createRewardAddress(): Promise<void> {
+    const mnemonic = mnemonicGenerate()
+    console.log(`Generated mnemonic: ${mnemonic}`)
```
nazar-pc, line 233: I wouldn't print mnemonic to the console

nazar-pc (commented):
Some minor comments.

I didn't read through all in much detail, but I think it makes sense overall.

The only thing I'm a little bit on a fence is to have `userConfirmed`. If we generate mnemonic client-side and already store reward address, missing reward address implies user didn't confirm they recorded mnemonic. So if app doesn't store reward address until user confirms mnemonic (we can generate it right when they clear reveal button) we don't need additional boolean flag.

src-tauri/src/main.rs
ozgunozerk: unfortunately yes, the conversion for the publicAddress format conversion on the frontend side requires localApi.
It sounds bad, I know. But no worries, after this PR merges, this whole address thing needs to be revized in the multi-farming setting (each plot having a public address). That's my next todo, so I vote for leaving this as is atm.

ozgunozerk:
`userConfirmed` is essentially `userConfirm` in the `SaveKeys.vue`. 
Its functionality is to force the user to copy the key (prevent the user to click `reveal` and proceeding in a hurry without actually securely storing the mnemonic).

We can generate a new one when the user clicks on reveal, that's true. But I didn't quite understand how we can ensure the user **copied** the mnemonic without having this boolean. Because the user copying the mnemonic to an outside environment is not trackable from the app, maybe he wrote the mnemonic on a paper?

nazar-pc (dismissed):

nazar-pc (approved):

PR #167, opened 2022-04-19
address structure is fixed

ozgunozerk:
This PR revamps the address structure (publicKey and rewardAddress). 

Changes:
- the farmer returns the publicKey instead of the node
- and since we unified the workflow of providing a reward address always to the farmer, `farming` function is also simplified.
- Node does not return anything anymore. 
- the publicKey address formatting on the frontend is simplified.
- reward address related optimizations on frontend

### DIFF
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -43,28 +43,19 @@ fn plot_progress_tracker() -> usize {
 }
 
 #[tauri::command]
-async fn farming(path: String, reward_address: String, plot_size: u64) {
-    match reward_address.len() {
-        0 => {
-            farm(path.into(), "ws://127.0.0.1:9944", None, plot_size)
-                .await
-                .unwrap();
-        }
-        _ => {
-            if let Ok(address) = parse_reward_address(&reward_address) {
-                farm(path.into(), "ws://127.0.0.1:9944", Some(address), plot_size)
-                    .await
-                    .unwrap();
-            } else {
-                error!("Reward address could not be parsed!");
-            }
-        }
+async fn farming(path: String, reward_address: String, plot_size: u64) -> [u8; 32] {
+    if let Ok(address) = parse_reward_address(&reward_address) {
+        farm(path.into(), "ws://127.0.0.1:9944", Some(address), plot_size)
+            .await
+            .unwrap()
+    } else {
+        unreachable!("Reward address could not be parsed on the backend!")
     }
 }
 
 #[tauri::command]
-async fn start_node(path: String) -> [u8; 32] {
-    init_node(path.into()).await.unwrap()
+async fn start_node(path: String) {
+    init_node(path.into()).await.unwrap();
 }
 
 #[tauri::command]
@@ -164,10 +155,7 @@ async fn main() -> Result<()> {
     Ok(())
 }
 
-async fn init_node(base_directory: PathBuf) -> Result<[u8; 32]> {
-    let identity = Identity::open_or_create(&base_directory)?;
-    let public_key = identity.public_key().to_bytes();
-
+async fn init_node(base_directory: PathBuf) -> Result<()> {
     let chain_spec =
         sc_service::GenericChainSpec::<subspace_runtime::GenesisConfig>::from_json_bytes(
             include_bytes!("../chain-spec.json").as_ref(),
@@ -188,7 +176,7 @@ async fn init_node(base_directory: PathBuf) -> Result<[u8; 32]> {
         }
     });
 
-    Ok(public_key)
+    Ok(())
 }
 
 /// Start farming by using plot in specified path and connecting to WebSocket server at specified
@@ -198,7 +186,7 @@ async fn farm(
     node_rpc_url: &str,
     reward_address: Option<PublicKey>,
     plot_size: u64,
-) -> Result<()> {
+) -> Result<[u8; 32]> {
     let identity = Identity::open_or_create(&base_directory)?;
     let address = identity.public_key().to_bytes().into();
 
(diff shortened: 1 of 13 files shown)


src-tauri/src/main.rs
```diff
@@ -43,28 +43,19 @@ fn plot_progress_tracker() -> usize {
...
-        _ => {
-            if let Ok(address) = parse_reward_address(&reward_address) {
-                farm(path.into(), "ws://127.0.0.1:9944", Some(address), plot_size)
-                    .await
-                    .unwrap();
-            } else {
-                error!("Reward address could not be parsed!");
-            }
-        }
+async fn farming(path: String, reward_address: String, plot_size: u64) -> [u8; 32] {
```
nazar-pc, line 46: Why does it still return something?

src-tauri/src/main.rs
```diff
@@ -43,28 +43,19 @@ fn plot_progress_tracker() -> usize {
...
+    if let Ok(address) = parse_reward_address(&reward_address) {
+        farm(path.into(), "ws://127.0.0.1:9944", Some(address), plot_size)
+            .await
+            .unwrap()
+    } else {
+        unreachable!("Reward address could not be parsed on the backend!")
```
nazar-pc, line 52: `unreachable!()` is for code that never happens. This should have returned an error to the frontend instead (or at least boolean about whether operation succeeded).

src/lib/client.ts
```diff
@@ -86,8 +85,10 @@ export class Client {
                   // TODO
                 }
               })
-              const addr: string | null = LocalStorage.getItem("rewardAddress")
-              const addr2: string = addr ?? farmerPublicKey
+              const addr_unchecked: string | null = LocalStorage.getItem("rewardAddress")
+              let addr!: string;
```
nazar-pc, line 89: You're just shutting up TypeScript compiler while showing at the same time that you're doing it incorrectly because `addr_unchecked` clearly may or may not be there, so `addr` is not guaranteed to be `string`.

src/lib/client.ts
```diff
@@ -86,8 +85,10 @@ export class Client {
...
+              if (addr_unchecked) addr = addr_unchecked
+              else console.log("Error: Reward Address should not be null!")
```
nazar-pc, line 91: If you have an error, use `console.error`, not `console.log`

src/pages/PlottingProgress.vue
```diff
@@ -235,7 +219,9 @@ export default defineComponent({
       const config = appConfig.getAppConfig()
       if (config) {
         await this.client.startBlockSubscription()
-        await this.client.startFarming(this.plotDirectory, config.plot.sizeGB)
+        const publicKey = await this.client.startFarming(this.plotDirectory, config.plot.sizeGB)
+        if (publicKey) appConfig.updateAppConfig(null, { farmerPublicKey: publicKey.toString() }, null, null, null, true)
+        else console.log("NO PUBLIC KEY RETURNED FROM FARMER")
```
nazar-pc, line 224: Same here, `console.error`

src/lib/client.ts
nazar-pc: Also for readability please always use `if {} else {}` even if you have single instruction in it.

src-tauri/src/main.rs
ozgunozerk: at the time, multi-replica is not merged to main. So, we need the public address from the backend to be able distinguish which blocks are farmed by us.

src-tauri/src/main.rs
ozgunozerk: Actually this should never happen. It is a double-check. 
We are already parsing the reward address on the frontend side and ensuring it is in the correct format. 
I can as well remove the double-check on the backend side if you like

src-tauri/src/main.rs
ozgunozerk: So, this will not return anything in the multi-replica PR :)

nazar-pc: How is multi-replica related to this if reward address is supported since initial March release?

src-tauri/src/main.rs
nazar-pc: I mean from the point of view of the backend it can definitely happen. You can't make that decision based on locally available information, hence `unreachable!()` is not appropriate.

src-tauri/src/main.rs
ozgunozerk: For reward address, we need to return the farmer's public address to the frontend (to see which blocks are farmed by the farmer). 

Here is how multi-replica steps in: since there can be multiple plots, this function will return multiple public keys (and the received block will be checked against all of them).

My previous comment was wrong (i was confused), it will return `Vec<[u8; 32]>` in multi-replica PR

src-tauri/src/main.rs
ozgunozerk: Ok, makes sense

src-tauri/src/main.rs
nazar-pc: I don't get why it needs to return anything either in multi or single replica. Just send reward address from frontend and you will always have the same reward address in produced blocks. Frontend doesn't need to know anything about backend identities.

ozgunozerk: Hmm maybe its a deeper issue. 
Leo designed the frontend as such: it is listening to the new produced blocks. Then compares the author of these new produced block with the farmer's public key. 

Instead of comparing the authors, we can compare the reward address embedded in the blocks, is this what you are suggesting?

nazar-pc: Yes.

Thinking more about it, the only challenge with that is if there is another farmer producing blocks for the same reward address, then it will show blocks produced by another farmer, which may or may not be desirable.

ozgunozerk: I also thought about that but the application is for displaying the rewards in the end. So I think it's not a problem. Will do the changes in the frontend for that

nazar-pc (approved):

PR #168, opened 2022-04-19
Multi replica

ozgunozerk:
Creating as a draft PR at the moment. 
When we do the runtime upgrade on the monorepo side, I will remove the custom chainspec file from here, and turn it into a real PR

### DIFF
--- a/src-tauri/src/farmer.rs
+++ b/src-tauri/src/farmer.rs
@@ -6,7 +6,7 @@ use std::sync::Arc;
 use std::time::Duration;
 use subspace_core_primitives::{PublicKey, PIECE_SIZE};
 use subspace_farmer::multi_farming::MultiFarming;
-use subspace_farmer::{Identity, ObjectMappings, RpcClient, WsRpc};
+use subspace_farmer::{Identity, NodeRpcClient, ObjectMappings, RpcClient};
 use subspace_rpc_primitives::FarmerMetadata;
 
 pub(crate) static PLOTTED_PIECES: AtomicUsize = AtomicUsize::new(0);
@@ -28,7 +28,7 @@ pub(crate) async fn farm(
     };
 
     info!("Connecting to node at {}", node_rpc_url);
-    let client = WsRpc::new(node_rpc_url).await?;
+    let client = NodeRpcClient::new(node_rpc_url).await?;
 
     let FarmerMetadata { max_plot_size, .. } = client
         .farmer_metadata()
--- a/src-tauri/src/farmer.rs
+++ b/src-tauri/src/farmer.rs
@@ -30,12 +30,7 @@ pub(crate) async fn farm(
     info!("Connecting to node at {}", node_rpc_url);
     let client = WsRpc::new(node_rpc_url).await?;
 
-    let FarmerMetadata {
-        record_size: _,
-        recorded_history_segment_size: _,
-        max_plot_size,
-        ..
-    } = client
+    let FarmerMetadata { max_plot_size, .. } = client
         .farmer_metadata()
         .await
         .map_err(|error| anyhow!(error))?;
@@ -84,7 +79,9 @@ pub(crate) async fn farm(
         }))
         .detach();
 
-    multi_farming.wait().await
+    tokio::spawn(async { multi_farming.wait().await });
+
+    Ok(())
 }
 
 pub(crate) fn parse_reward_address(s: &str) -> Result<PublicKey, sp_core::crypto::PublicError> {
--- a/src-tauri/src/farmer.rs
+++ b/src-tauri/src/farmer.rs
@@ -1,0 +1,128 @@
+use anyhow::Result;
+use bip39::{Language, Mnemonic};
+use log::{debug, error, info};
+use std::path::PathBuf;
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::Arc;
+use std::time::Duration;
+use subspace_core_primitives::{PublicKey, PIECE_SIZE};
+use subspace_farmer::{
+    Commitments, FarmerData, Farming, Identity, ObjectMappings, Plot, Plotting, RpcClient, WsRpc,
+};
+use subspace_solving::SubspaceCodec;
+
+pub(crate) static PLOTTED_PIECES: AtomicUsize = AtomicUsize::new(0);
+const BEST_BLOCK_NUMBER_CHECK_INTERVAL: Duration = Duration::from_secs(5);
+
+pub(crate) fn create_identity(base_directory: PathBuf) -> Result<String> {
+    let identity = Identity::open_or_create(&base_directory)?;
+
+    Ok(
+        Mnemonic::from_entropy(identity.entropy(), Language::English)
+            .unwrap()
+            .into_phrase(),
+    )
+}
+
+/// Start farming by using plot in specified path and connecting to WebSocket server at specified
+/// address.
+pub(crate) async fn farm(
+    base_directory: PathBuf,
+    node_rpc_url: &str,
+    reward_address: Option<PublicKey>,
+    plot_size: u64,
+) -> Result<()> {
+    let identity = Identity::open_or_create(&base_directory)?;
+    let address = identity.public_key().to_bytes().into();
+
+    let reward_address = reward_address.unwrap_or_else(|| identity.public_key().to_bytes().into());
+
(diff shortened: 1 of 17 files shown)


src/loc/en.json
```diff
@@ -30,6 +30,7 @@
     "utilizedSpace": "How much space is currently being used.",
     "allocatedSpace": "Space to be shared with Subspace Protocol.",
     "estimatingSpace": "Requesting the blockchain size from the network.",
+    "nodeSpace": "Space Requied for Node",
```
ozgunozerk, line 33: note to myself: also get rid of this when deleting the custom chain-spec file

src-tauri/src/farmer.rs
```diff
@@ -31,52 +17,27 @@ pub(crate) async fn farm(
...
 
     info!("Connecting to node at {}", node_rpc_url);
-    let client = WsRpc::new(node_rpc_url).await?;
-
-    let farmer_metadata = client
+    let client = WsRpc::new(&node_rpc_url).await?;
+
+    let FarmerMetadata {
+        record_size: _,
+        recorded_history_segment_size: _,
```
i1i1, line 35: Should those 2 lines be removed?

src-tauri/src/main.rs
```diff
@@ -36,15 +39,27 @@ fn plot_progress_tracker() -> usize {
 async fn farming(path: String, reward_address: String, plot_size: u64) {
```
i1i1: Btw, does tauri support `Option<String>` and `PathBuf` for arguments? It would be handy to use those here

ozgunozerk: unfortunately no :/ 
All the parameters and return types for `tauri command` functions need to be primitive types, since these are communicating with frontend.

i1i1: I believe `Option<String>` at least should be? There is `null` in js

src-tauri/src/farmer.rs
ozgunozerk: agreed, thanks!

src-tauri/src/main.rs
ozgunozerk: Maybe, but I think it serves no purpose for desktop app. Since there **always** will be `reward_address` and `path`. 

The reason for mandatory reward address: 
the first case: if the user decides to import reward address -> user enters the reward address, and frontend sends the reward address to backend.
the second case: user does not import reward address -> because of multi-replica, there has to be a ROOT-IDENTITY for reward address. We choose to generate this new identity on the frontend, for unified workflow (frontend sending the reward address to backend). 

In summary, there is no path in our workflow to *NOT* send a reward address to backend.

i1i1: Ok, should we delete branch there we check for `length == 0` on reward address?

i1i1: Missed changes in latter commits. It is no longer needed.

i1i1 (approved):

ozgunozerk:
Please give it a final look, it is ready for a real review now :)

src-tauri/src/node.rs
```diff
@@ -148,63 +173,66 @@ fn create_configuration<CS: ChainSpec + 'static>(
...
-        tracing_targets: None,
-        tracing_receiver: TracingReceiver::Log,
-        chain_spec: Box::new(chain_spec),
-        max_runtime_instances: 8,
-        announce_block: true,
-        role,
-        base_path: Some(base_path),
-        informant_output_format: Default::default(),
-        runtime_cache_size: 2,
+        force_new_slot_notifications: false,
```
ozgunozerk, line 235: Nazar, I don't not 100% sure this should be false/true. Can you double check?

src/lib/client.ts
```diff
@@ -53,26 +54,22 @@ export class Client {
       },
       start: async (): Promise<void> => {
 
-        const farmerAddress = LocalStorage.getItem("rewardAddress")
-        if (farmerAddress === null) {
-          console.error("Reward address should not have been null...")
+        const farmerAddress: string | undefined = appConfig.getAppConfig()?.rewardAddress
+        if (farmerAddress == null) {
```
isSerge, line 58: `===` is usually preferred over `==`
in this case, you can just do `if (!farmerAddress) {`

src/pages/SetupPlot.vue
```diff
@@ -299,10 +295,16 @@ export default defineComponent({
     },
     async checkIdentity() {
       const config = appConfig.getAppConfig()
-      if (config && config.importedRewAddr === false) {
-        await this.client.createRewardAddress()
-        await this.viewMnemonic()
+      if (config) {
+        if (config.rewardAddress == "") {
```
isSerge, line 299: `if (!config.rewardAddress)` is better

src-tauri/src/node.rs
ozgunozerk: forgot to tag properly in the previous comment -> @nazar-pc

nazar-pc: `false` is fine

src-tauri/src/node.rs
```diff
@@ -148,63 +173,75 @@ fn create_configuration<CS: ChainSpec + 'static>(
...
+                other: ExecutionStrategy::AlwaysWasm,
+            },
+            #[cfg(target_os = "windows")]
+            execution_strategies: ExecutionStrategies {
+                syncing: ExecutionStrategy::NativeElseWasm,
+                importing: ExecutionStrategy::NativeElseWasm,
+                block_construction: ExecutionStrategy::NativeElseWasm,
+                offchain_worker: ExecutionStrategy::NativeElseWasm,
+                other: ExecutionStrategy::NativeElseWasm,
+            },
```
ozgunozerk, line 208: these are the latest changes from release preparations (monorepo) @nazar-pc

src-tauri/src/node.rs
```diff
@@ -148,63 +173,75 @@ fn create_configuration<CS: ChainSpec + 'static>(
...
+            network,
+            keystore_remote,
+            keystore,
+            database: database_config(&config_dir, database_cache_size, &role),
+            state_cache_size: 67_108_864,
+            state_cache_child_ratio: None,
+            // TODO: Change to constrained eventually (need DSN for this)
+            state_pruning: PruningMode::keep_blocks(1024),
+            keep_blocks: KeepBlocks::Some(1024),
+            wasm_method: WasmExecutionMethod::Interpreted,
```
nazar-pc, line 191: Ouch, no interpreted WASM!!!

src/pages/SetupPlot.vue
```diff
@@ -173,7 +172,7 @@ export default defineComponent({
     },
     stats(): StatsType {
       const totalDiskSizeGB = util.toFixed(this.driveStats.totalBytes / 1e9, 2)
-      const safeAvailableGB = this.driveStats.freeBytes / 1e9
+      const safeAvailableGB = this.driveStats.freeBytes / 1e9 - 20
```
nazar-pc, line 175: What is this 20? We need to allocate 10GB for node and the rest for farmer, but the user should only specify the combined number, they only care how much will software use in total.

ozgunozerk: We have talked with you previously that node is at most occupy 20GB. 
So I put 20GB for a safe cap. 

Jeremiah and I discussed that we should not complicate the user workflow and maybe should not even mention about that (node occupying space). So user can enter the plot size, and we can deduct 20GB from the safe space.

I can update this to 10 if you like, no problem. 
However, when we switch to full-node, will it still occupy around 10GB? I think it would be around 1-2GB at most? Am I recalling it wrong?

src-tauri/src/node.rs
ozgunozerk: Since we removed `wasm exeuction compiled` from readme
https://github.com/subspace/subspace/pull/405/files#diff-cb44c4563e0dbee8cb7ed4e05a25e2708bd167ef5b59cabc8e9fcad2dbb70e73L39

I thought it would use the default.

and the default for `WasmExecutionMethod` was ->
```rust
impl Default for WasmExecutionMethod {
	fn default() -> WasmExecutionMethod {
		WasmExecutionMethod::Interpreted
	}
}
```

So we are using `compiled` and it is declared in somewhere else?

nazar-pc (approved):

nazar-pc (approved):

PR #178, opened 2022-04-25
disabled reset button on prod

ozgunozerk:
Disables the `reset` button from the main menu (which removes everything and basically performs a soft uninstall-reinstall) for the production mode. And keeps it only for the dev mode.

The motivation behind this is: and accidental click on this button (although there is a confirmation screen) can be very upsetting for the end users, and it serves more purpose for the dev env.

### DIFF
--- a/src/components/mainMenu.vue
+++ b/src/components/mainMenu.vue
@@ -12,7 +12,7 @@ q-menu(auto-close)
         .col
           p.text-grey(v-if="!launchOnStart") {{ lang.autoStart }}
           p.text-black(v-else) {{ lang.autoStart }}
-    q-item(@click="reset()" clickable)
+    q-item(@click="reset()" clickable v-if="devMode")
       .row.items-center
         .col-auto.q-mr-md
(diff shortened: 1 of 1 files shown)


src/components/mainMenu.vue
```diff
@@ -34,13 +34,19 @@ export default defineComponent({
...
+      devMode: false
     }
   },
   mounted() {
+    this.checkDev()
     this.initMenu()
   },
   methods: {
+    checkDev() {
+      if (util.CONTEXT_MENU != "OFF")
```
isSerge, line 47: do we use `devMode` anywhere else? I guess we can just use `util.CONTEXT_MENU !== "OFF"` in `v-if=` and skip addiotional method and state

ozgunozerk: sure

ozgunozerk: no can do, since util is defined below vue part

isSerge (approved):

PR #180, opened 2022-04-25
display node name and version

ozgunozerk:
Here is how it looks:
image

Please check the notes on frontend below

### DIFF
--- a/src/layouts/MainLayout.vue
+++ b/src/layouts/MainLayout.vue
@@ -5,13 +5,15 @@ q-layout(view="hHh lpr fFf")
       q-img(no-spinner src="subspacelogo.png" width="150px")
       q-toolbar-title
         .row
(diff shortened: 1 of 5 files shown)


src/layouts/MainLayout.vue
```diff
@@ -58,10 +68,21 @@ export default defineComponent({
...
     }
   },
-
-  methods: {}
+  mounted() {
+    this.nodeNameChanger()
+  },
+  methods: {
+    async nodeNameChanger() {
+      let _this = this
```
ozgunozerk, line 80: i cheated by using `_this`. 
maybe fat arrows may solve the `this scope` issue.

nazar-pc: Yes, use fat arrows instead.

isSerge (dismissed):

isSerge (approved):

PR #189, opened 2022-04-29
dynamically display version in MainLayout

ozgunozerk:


src/layouts/MainLayout.vue
```diff
@@ -55,6 +55,7 @@ import { globalState as global } from "src/lib/global"
 import * as util from "src/lib/util"
 import MainMenu from "components/mainMenu.vue"
 import { myEmitter } from "src/lib/client"
+import * as process from 'process'
```
isSerge, line 58: it does not work without this import?

ozgunozerk: you are right, it does work without it!

ozgunozerk: btw since `process.env.APP_VERSION` is not actually a string, but `string | undefined`, it's not beautiful to use it directly in the vue and creates problems. Hence the extra variable `appVersion`.

ozgunozerk: weird, it does work in dev mode (restarted server as well), but `build` script returns error about `process` undefined. So it seems we need it

isSerge: you can do type assetion like this:
```
const nodeEnv: string = (process.env.NODE_ENV as string);
```

ozgunozerk: thats great, thanks

isSerge (approved):

PR #190, opened 2022-05-02
Readme os specs

ozgunozerk:
First iteration for the README -> OS specs. 
Nothing fancy, but feel free to improve/better it

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,14 @@
 Standalone Desktop application for farming on Subspace Network (includes both node and farmer).
 Download and run this application to be a farmer on Subspace Network (currently on testnet).
 
+# Minimum OS Requirements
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -3,6 +3,14 @@
 Standalone Desktop application for farming on Subspace Network (includes both node and farmer).
 Download and run this application to be a farmer on Subspace Network (currently on testnet).
 
+# Minimum OS Requirements
```
nazar-pc, line 6: I think we should rather have supported OSs explicitly. For instance I'm not sure 18.04 will work properly and it is old enough to not bother about.

I'd like Windows 10 and 11, Ubuntu 20.04 and 22.04, macOS 11 and 12. Just don't claim supporting anything else unless we actually confirmed them working and are willing to support.

PR #192, opened 2022-05-04
External config

ozgunozerk:
This PR allows us to store the config information in an external file, which will be persistent across different versions.
With this change, we will be able to store the following information:
- reward address
- existing plot location and size
- version number (of the app)
- launchOnBoot preference

This PR resolves a very important bug: *when a newer version is installed, the app was unable to detect the existing plot from the older version*. 

Also, this change will potentially enable us to detect breaking changes in the code and take actions accordingly throughout the update process.
And lastly, all the other information will also be remembered across different versions, which is a huge improvement on the UX side.

Other small improvements:
- `disable()` function for autoLauncher is improved -> if there are multiple entries for `subspace-desktop` (which can occur with different version installations), now the `disable()` functions removes all of them.
- there were some `console.error`s present in the code, which were not actually errors. These were polluting the console when debugging, and also will be polluting the log when we have an external log file in the future. So I pruned and handled these unnecessary `console.error`s along the way.
- some minor refactorings for better structure


nazar-pc:
I'd prefer for Serge to review this, let me know if my input is necessary

src/lib/appConfig.ts
```diff
@@ -1,31 +1,105 @@
...
+
+  // TODO: remove this after PlottingProgress refactoring
+  segmentCache: SegmentCache,
+}
+
+const emptyConfig: Config = {
+  plot: { location: "", sizeGB: 0 },
+  rewardAddress: "",
+  launchOnBoot: true,
+  version: process.env.APP_VERSION || "?",
```
isSerge, line 30: don't we assume `APP_VERSION` is always a string? we never display `"?"`

src/lib/autoLauncher.ts
```diff
@@ -150,22 +150,25 @@ export class AutoLauncher {
     if (!this.enabled) {
       console.error("ENABLE DID NOT WORK")
     } else {
-      appConfig.updateAppConfig(null, null, true, null, null)
+      await appConfig.update(null, null, true, null, null)
```
isSerge, line 153: not critical, but maybe it will be more readable to pass an object, and instead of `null`s we just provide values for properties we need. In this case, it would be:
```
{ launchOnBoot: true }
```

src/lib/autoLauncher.ts
```diff
@@ -181,16 +184,15 @@ export class AutoLauncher {
...
-      }
+    const config = await appConfig.read()
+    if (config.launchOnBoot)
+    {
+      // the app may be initialized before, but then user may have decided to move the app to another directory
+      // in this case, we have to delete the previous autoLaunch entry, and create a new one
+      // below disable is not creating console error, hence use it for this one
+      await this.disable()
+      await this.enable()
+      this.enabled = (await this.isEnabled())
```
isSerge, line 195: this part is a bit confusing:
Does not `this.enable()` update `this.enabled` property? (line 149)

src/lib/appConfig.ts
```diff
@@ -1,31 +1,105 @@
...
-      if (plottingStarted != null) newAppConfig.plottingStarted = plottingStarted
-      LocalStorage.set("appConfig", newAppConfig)
-    }
+    launchOnBoot: boolean | null,
+    version: string | null,
+    segmentCache: SegmentCache | null,
+  ): Promise<void> {
+    const newAppConfig = await this.read()
+    if (plot) newAppConfig.plot = plot
+    if (launchOnBoot != null) newAppConfig.launchOnBoot = launchOnBoot
```
isSerge, line 95: why not `if (launchOnBoot)`?

src/lib/appConfig.ts
ozgunozerk: the reason is, `process.env.APP_VERSION` return type is `string | null` or `string | undefined`. 
If the `APP_VERSION` cannot be resolved, there must be something wrong. That's why I put the "?", and made `version` accept type of `string` only.

src/lib/appConfig.ts
ozgunozerk: that was a really funny bug. 
Imagine `launchOnBoot` is `false`. 
Then the following line:
`if (launchOnBoot) newAppConfig.launchOnBoot = launchOnBoot`
won't execute, because if statement will evaluate to false :D

src/lib/appConfig.ts
isSerge: can do `process.env.APP_VERSION as string`

src/lib/appConfig.ts
isSerge: makes sense, `launchOnBoot !== null` is preferred then. I will adjust eslint rules later

src/lib/autoLauncher.ts
ozgunozerk: wonderful! implementing now

src/lib/autoLauncher.ts
ozgunozerk: true, nice catch 👍

src/lib/appConfig.ts
```diff
@@ -1,31 +1,113 @@
...
+  },
+  async update(
+    {
+      plot = null,
+      launchOnBoot = null,
+      rewardAddress = null,
+      version = null,
+      segmentCache = null,
+    }: {
+      plot?: Plot | null;
```
isSerge, line 94: I suspect it should work without `= null` and `| null`

isSerge: if we don't provide value for prop explicitly it will be `undefined`

isSerge: but also update this `if (launchOnBoot != null)`

isSerge (dismissed):

isSerge (approved):

PR #196, opened 2022-05-06
upgraded tauri, and removed the round icon from mainlayout

ozgunozerk:
Upgraded tauri to latest version (required for enabling the logger plugin they have).
The next PR will be about external log file mechanism.

Along the way, fixed a small issue: removed the round icon from `MainLayout`. 
The main reason for this is: when `node name` is too long, this round icon overflows to the next line, making the UI very ugly.
The second reason is: it is not very clear what does this icon do from the end-user perspective, and the application will be more minimalistic without it.
image

### DIFF
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -16,10 +16,11 @@ use serde::Serialize;
 use std::path::PathBuf;
 use std::sync::atomic::Ordering;
 use std::time::Duration;
+
 use tauri::SystemTrayEvent;
 use tauri::{
     api::{self},
-    Env, Manager, RunEvent,
+    Env, Manager, RunEvent, WindowEvent,
 };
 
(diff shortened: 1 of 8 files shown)


nazar-pc:
Looks like it doesn't compile on Linux yet

ozgunozerk:
Yes, they removed `ayatana-tray` from default. We can choose between `gtk-tray` and `ayatana-tray`. 
Do you have any preference?
I'm researching about them currently

src-tauri/src/main.rs
```diff
@@ -16,10 +16,11 @@ use serde::Serialize;
 use std::path::PathBuf;
 use std::sync::atomic::Ordering;
 use std::time::Duration;
+
```
nazar-pc, line 19: Please no new line

nazar-pc (approved):

nazar-pc (dismissed):

nazar-pc (approved):

PR #198, opened 2022-05-09
Logging and small fixes

ozgunozerk:
trivial logging architecture for logging.

useful information for reviewing this PR:
- it is using a native plugin from tauri team for logging: https://github.com/tauri-apps/tauri-plugin-log
- backend exposes two new functions for logging to the frontend (1 for error, 1 for info level)
- frontend uses these exposed functions to relay the error/log messages to backend
- tauri handles all the process of writing logs to an external file
- external log file location is not configurable from the end-user perspective. We are using the default ones for macos, linux and windows respectively (actually, tauri-plugin-log is handling these)

I suggest reviewing commit by commit

P.S. 
I changed the identifier name from `com.tauri.dev` to `com.subspace.desktop`

### DIFF
--- a/src-tauri/src/node.rs
+++ b/src-tauri/src/node.rs
@@ -13,7 +13,6 @@ use sc_service::{
     BasePath, Configuration, DatabaseSource, KeepBlocks, PruningMode, Role, RpcMethods,
     TracingReceiver,
 };
-use sc_tracing::logging::LoggerBuilder;
 use sp_core::crypto::Ss58AddressFormat;
 use std::env;
 use std::path::{Path, PathBuf};
@@ -93,10 +92,6 @@ pub(crate) async fn create_full_client<CS: ChainSpec + 'static>(
             "https://discord.gg/vhKF9w3x",
             env!("SUBSTRATE_CLI_IMPL_VERSION"),
         );
-
-        LoggerBuilder::new("info")
-            .init()
-            .expect("Logger initialization must not fail");
 
         if let Some(new_limit) = fdlimit::raise_fd_limit() {
             if new_limit < RECOMMENDED_OPEN_FILE_DESCRIPTOR_LIMIT {
--- a/src/lib/autoLauncher.ts
+++ b/src/lib/autoLauncher.ts
@@ -6,6 +6,7 @@ import * as fs from "@tauri-apps/api/fs"
 import * as native from './native'
 import * as path from "@tauri-apps/api/path"
 import { appConfig } from "src/lib/appConfig"
+import { errorLogger, infoLogger } from "./util"
 
 type osAL = typeof macAL | typeof winAL | typeof linAL | typeof nullAL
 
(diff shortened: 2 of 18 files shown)


src/lib/autoLauncher.ts
```diff
@@ -38,24 +39,30 @@ Comment=${appName} startup script
...
     try {
       const autostartAppFile = await this.getAutostartFilePath(appName)
-      await fs.readTextFile(autostartAppFile).catch(console.error)
+      await fs.readTextFile(autostartAppFile).catch((error) => {
+        errorLogger(error)
+      })
       return true
     } catch (error) {
-      console.error(error)
+      errorLogger(error as string)
```
isSerge, line 65: error can be both `string` and instance of `Error`, so should be something like:
```
if (typeof error === "string") {
        errorLogger(error)
    } else if (error instanceof Error) {
        errorLogger(error.message)
    }
    ```

Comment (isSerge) [reply src/lib/autoLauncher.ts:65]: same in other places when you catch errors

Comment (ozgunozerk) [reply src/lib/autoLauncher.ts:65]: oh ok thanks!

File: src-tauri/src/node.rs [hunk 2]
```diff
@@ -13,7 +13,6 @@ use sc_service::{
     BasePath, Configuration, DatabaseSource, KeepBlocks, PruningMode, Role, RpcMethods,
     TracingReceiver,
 };
-use sc_tracing::logging::LoggerBuilder;
```
nazar-pc, line 16: So we can remove dependency?

src/pages/Dashboard.vue
```diff
@@ -80,11 +80,14 @@ export default defineComponent({
     const raceResult = util.promiseTimeout(7000, this.client.connectPublicApi())
     raceResult.then(async() => {
       if (this.client.isFirstLoad() === false) {
+        console.log("will start node now ***********")
```
nazar-pc: In IntelliJ you can commit part of the changes. No need to commit these debugging stuff.
You can also create change sets, just move debugging stuff into inactive change set and it will not be selected for committing in the first place.

src/pages/SetupPlot.vue
```diff
@@ -252,19 +253,22 @@ export default defineComponent({
...
       if (dirExists) {
+        console.log("SETUP PLOT | found the old plotting directory")
         const files = await tauri.fs
           .readDir(this.plotDirectory)
           .catch(console.error)
 
         if (files) {
           console.log("FILES ARE: :", files)
-          if (files.length === 0 || !files.some(item => item.name === "chains" )) {
+          if (files.length === 0 ||  files.length === 1 && files.some(item => item.name === "subspace-desktop.cfg")){
```
nazar-pc, line 265: Please never ever use `a || b && c`, I don't have to remember what the priority of operators should be in this case in a specific language.

src/lib/util.ts
```diff
@@ -7,6 +7,9 @@ import { Component } from "vue"
 import * as process from "process"
 import { appData } from "./appData"
 import { appConfig } from "./appConfig"
+import { invoke } from "@tauri-apps/api/tauri"
+
+const tauri = { invoke }
```
nazar-pc, line 12: Why though?

src/pages/Dashboard.vue
ozgunozerk: Thanks for the suggestion!

src-tauri/src/node.rs
ozgunozerk: it is already deleted

isSerge (approved):

isSerge (dismissed):

src-tauri/src/node.rs
nazar-pc: Are you sure you're looking at the correct revision? I can still clearly see it in `Cargo.lock`.

ozgunozerk: I just checked out at this branch from scratch. No `sc_tracing` in the repo. Also no `LoggerBuilder` aside from `tauri_plugin_log::LoggerBuilder`. 
I think we are good, but will not merge until you confirm too

nazar-pc: Some on, open "Files changed" and look at line 27 of `Cargo.toml`...

nazar-pc (approved):

PR #205, opened 2022-05-12
contributing and code_of_conduct

ozgunozerk:
Please take a look, 

I think this can be the first iteration of our contribution guideline.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,28 @@
+The Subspace community is big on decentralization. That is why, it is an open-source project, and are trying our best to be as transparent as possible. To promote Subspace Network being a community driven project, we take any breaches of mutual respect very seriously. Below is the community code of conduct, which applies to our Github repos and organizations, Discord and Telegram channels, and any other Subspace-supported communication group.
+
+### Be excellent to each other.
+
+Treat everyone with respect. Participate while acknowledging this is an open-source project and that everyone deserves to be here — and each of us has the right to enjoy our experience and participate without fear of harassment, discrimination, or condescension, whether blatant or via micro-aggressions.
+
+### Speak up if you see or hear something.
+
+Harassment is not tolerated, and you are empowered to politely engage when you or others are disrespected. Also, assume positive intent. The person making you feel uncomfortable may not be aware of what they are doing, and politely bringing their behavior to their attention is encouraged.
+
+### Practice saying "Yes, and" to each other instead of "No".
+
+Build on each other’s ideas. We all benefit when we create together.
+
+### We have a ZERO TOLERANCE POLICY for harassment of any kind, including but not limited to:
+
+- Stalking/following
+- Harassing comments
+- Deliberate intimidation
+- Sustained disruption of threads, channels, lists, etc.
+- Offensive comments
+- Sexual images in public spaces
+- Unwelcome sexual or otherwise aggressive attention
+- and any inappropriate content that is not listed here (according to common-sense)
+
+We empower and encourage you to report any behavior that makes you or others feel uncomfortable by emailing community@subspace.network.
+
+Any person communicating in a disorderly manner, engaging in harassing or uncomfortable behavior, or otherwise failing to comply with this policy may be removed from any Subspace-supported platform or permanently banned.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,74 @@
+## How to Create an Issue (first iteration):
+# ![Subspace](subspace-cover.jpg)
+
+## Contributing to Subspace
+
+Thank you for considering to help out with the source code! Subspace welcomes help in many forms including development, code review, documentation improvements, and outreach. We are grateful for even the smallest of improvements!
+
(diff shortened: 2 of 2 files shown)


ozgunozerk:
We should wait until 0.4.9 becomes stable for merging this PR, because for 0.3.3 we don't have logging.

Also before merging:
- [ ] let Skyler and Diana know about the new Issue Template

CONTRIBUTING.md
```diff
@@ -0,0 +1,74 @@
...
+
+*For complex implementations you are advised to first discuss the feature implementation or bug fix using an issue.*
+
+
+## Development
+
+### Code style
+
+Subspace is following the
+[Substrate code style](https://github.com/paritytech/substrate/blob/master/docs/STYLE_GUIDE.md).
```
nazar-pc, line 58: No, Subspace is most certainly not following that.

CONTRIBUTING.md
```diff
@@ -0,0 +1,74 @@
...
+We have a [Code of Conduct](./CODE_OF_CONDUCT.md), please follow it in all interactions with project maintainers and fellow users :)
+
+## Issues
+
+You want to write an issue if you have a
+- feature request
+- bug report
+- technical question
+
+### How to Create an Issue?
```
nazar-pc, line 19: These should be just GitHub templates if we want to have them at all.

CONTRIBUTING.md
```diff
@@ -0,0 +1,74 @@
...
+    - Submit the log file
+- If it is a feature request:
+    - Clearly explain the feature that is being requested
+    - Explain why do you think we need this feature / how would it enhance the application
+
+**Where is my log file?**
+Below are the `log` file locations (replace `alice` with your username)
+- windows: `C:\Users\Alice\AppData\Roaming\com.subspace.desktop`
+- macos: `/Users/Alice/Library/Logs/com.subspace.desktop`
+- linux: `/home/alice/.config/com.subspace.desktop`
```
nazar-pc, line 34: What are logs doing in `.config` on Linux? They should be in `.local/share` instead. `.config` is for configuration.

CONTRIBUTING.md
```diff
@@ -0,0 +1,74 @@
...
+## Development
+
+### Code style
+
+Subspace is following the
+[Substrate code style](https://github.com/paritytech/substrate/blob/master/docs/STYLE_GUIDE.md).
+
+In addition, we incorporate several tools to improve code quality. These are integrated into our CI and are expected to pass before a PR is considered mergeable. They can also be run locally.
+
+* [clippy](https://github.com/rust-lang/rust-clippy) - run with `cargo clippy --release --workspace`
```
nazar-pc, line 62: Is `--release` necessary here? It'll be faster without it.

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,28 @@
+The Subspace community is big on decentralization. That is why, it is an open-source project, and are trying our best to be as transparent as possible. To promote Subspace Network being a community driven project, we take any breaches of mutual respect very seriously. Below is the community code of conduct, which applies to our Github repos and organizations, Discord and Telegram channels, and any other Subspace-supported communication group.
```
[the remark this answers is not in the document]
ozgunozerk: I wrote this one couple of months ago. I don't exactly recall what I based this on :/
I can do a new research about this.

CONTRIBUTING.md
ozgunozerk: I'm using the default folder choices from `tauri-plugin-log`. Here it is:
https://github.com/tauri-apps/tauri-plugin-log/blob/dev/src/lib.rs#L106

I can open an issue on there to suggest `.local/share` instead.

ozgunozerk:
Turns out there are many things to change in here. And we want our users to use the forum instead of directly creating an issue. 
The main purpose of this PR was to provide a guideline for issue creation. 
I will prepare a github issue template instead. And we can return to this PR later. 
I believe @ImmaZoni is preparing something similar in parallel as well. Maybe he can use the information in here :)

Going to close this one for now. Maybe we can re-open it later

CONTRIBUTING.md
ozgunozerk: Btw, the convention for linux seems to be `var/log`. Is there a reason you prefer `.local/share` instead of `var/log`?

nazar-pc: Please do, I think it is really important to have logs in the correct location.

No, `/var/log` is for system stuff, regular users don't have access there, and even for system stuff it is generally preferred to go through things like systemd.

nazar-pc:
Yeah, IIRC Justin was preparing something, but can't find it right now. We'd ideally use something like that.

CODE_OF_CONDUCT.md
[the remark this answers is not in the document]
ImmaZoni: > I wrote this one couple of months ago. I don't exactly recall what I based this on :/ I can do a new research about this.

here is the off the shelf one we went with on the subspace docs as was discussed with Nazar a while ago https://github.com/subspace/subspace-docs/blob/main/.github/CODE_OF_CONDUCT.md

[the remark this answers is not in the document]
nazar-pc: Yes, I think we want that one everywhere

PR #207, opened 2022-05-13
Fix cannot read private member #isReadyPromise and dependencies update

isSerge:
- update dependencies
- fix cannot read private member #isReadyPromise
- adjust TS and Eslint

### DIFF
--- a/src/lib/client.ts
+++ b/src/lib/client.ts
@@ -6,7 +6,7 @@ import * as event from "@tauri-apps/api/event"
 import { invoke } from "@tauri-apps/api/tauri"
 import { reactive } from "vue"
 import * as process from "process"
-import * as util from "src/lib/util"
+import * as util from "../lib/util"
 import { appConfig } from "./appConfig"
 import { getStoredBlocks, storeBlocks } from "./blockStorage"
 import {
@@ -14,7 +14,7 @@ import {
   NetStatus,
   FarmedBlock,
   SubPreDigest
-} from "src/lib/types"
+} from "../lib/types"
 import EventEmitter from "events"
 
 export const myEmitter = new EventEmitter();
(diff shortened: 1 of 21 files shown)


src/lib/client.ts
```diff
@@ -26,24 +26,27 @@ const SUNIT = 1000000000000000000n
 const NETWORK_RPC = process.env.PUBLIC_API_WS || "ws://localhost:9944"
 const LOCAL_RPC = process.env.LOCAL_API_WS || "ws://localhost:9944"
 const appsLink = "https://polkadot.js.org/apps/?rpc=" + NETWORK_RPC + "#/explorer/query/"
+
+const publicApi = new ApiPromise({
```
nazar-pc, line 30: Why not doing this in constructor, so you have control over when it happens instead of running on file inclusion?

isSerge: I am afraid that is the actual fix for the problem. Originally API was assigned directly as a class property and it was causing the problem. Apparently, Quasar or Vue use `Proxy` somewhere under the hood - https://github.com/polkadot-js/api/issues/4656 Using constructor does not solve this.

nazar-pc: You need to use async contructor and wait for API to actually initialize before you use it, did you try that to make sure it is actually working?

If contructors are broken with Quasar or Vue then bad things will be literally everywhere, I don't believe that is the case, but async initialization of Polkadot.js's API might easily be.

isSerge: Async constructor may work, but it will require some refactoring (not just Client class), this was intended as a quick fix

nazar-pc: Can you at least not refactor things to use global variable and assign property of the class to the constant? And TODO there to refactor to async constructor.

isSerge: I made an attempt to fix this with as minimal effort as possible, but it requires much more. Please, check

src/lib/client.ts
```diff
@@ -1,45 +1,44 @@
...
-  protected localApi: ApiPromise = new ApiPromise({
-    provider: new WsProvider(LOCAL_RPC, false),
-    types: util.apiTypes
-  })
   protected farmed: FarmedBlock[] = []
   protected clearTauriDestroy: event.UnlistenFn = () => {}
   protected unsubscribe: event.UnlistenFn = () => {}
+
+  localApi: ApiPromise;
+  publicApi: ApiPromise;
```
nazar-pc, line 35: Should these be public?

tsconfig.json
```diff
@@ -4,7 +4,7 @@
   "compilerOptions": {
     "target": "ESNext",
     "strict": true,
-    "baseUrl": ".",
+    "baseUrl": "./src",
```
nazar-pc, line 7: I always wonder why is it `./src` and not just `src` :slightly_smiling_face:

nazar-pc (approved):
Generally looks better than initial attempt, but I am convinced over and over again that the architecture here isn't very reliable with all this global state.

isSerge:
@nazar-pc global state is usually a store like redux or vuex (similar for vue), never saw a state like this

src/lib/client.ts
isSerge: currently yes

nazar-pc (approved):

PR #208, opened 2022-05-13
Persistent node name

### LINKED ISSUE #197: Persist node name across restarts
Right now we just generate a random name on each restart, we need to at least make it persistent, maybe even customizable in the future.

ozgunozerk:
Fixes #197 

Also, refactors some frontend code such that the changed functions have more specific roles and they are better structured.

Lastly, `update` function for config was not `awaiting` on the `write` call to the filesystem. This was creating an edge case, where user closes the modal for viewing mnemonic (view mnemonic modal writes to config on dismiss). And after on dismiss, we are proceeding into `plottingProgress` page.
If the user closes the app immediately after proceeding into `plottingProgress` page, they will expect their reward address to be remembered. But the app sometimes may not have chance to write the `rewardAddress` to the config, since the write operation is not awaited on.
When the `write` operation is awaited, we will not proceed into `plottingProgress` page unless `rewardAddress` is correctly stored in config. 
Fixed also this 👍

### DIFF
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -12,7 +12,6 @@ mod node;
 
 use anyhow::Result;
 use log::{debug, error, info, LevelFilter};
-use node::generate_node_name;
 use serde::Serialize;
 use std::path::PathBuf;
 use std::sync::atomic::Ordering;
@@ -64,11 +63,6 @@ async fn farming(path: String, reward_address: String, plot_size: u64) -> bool {
         // reward address could not be parsed, and farmer did not start
         false
     }
-}
-
-#[tauri::command]
-fn init_node_name() -> String {
-    generate_node_name()
 }
 
 #[tauri::command]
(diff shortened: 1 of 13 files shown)


src-tauri/src/main.rs
```diff
@@ -66,12 +67,13 @@ async fn farming(path: String, reward_address: String, plot_size: u64) -> bool {
 }
 
 #[tauri::command]
-async fn start_node(path: String) -> String {
-    let node_name = node::generate_node_name();
-    node::init_node(path.into(), node_name.clone())
-        .await
-        .unwrap();
-    node_name
+fn init_node_name() -> String {
```
nazar-pc, line 70: Not a fan that this logic is forwarded through frontend. Don't we have a config on the backend? If not it would be preferred to move node name generation to the frontend completely I think instead of exposing `init_node_name` from the backend.

nazar-pc (approved):
This should work

src-tauri/src/main.rs
ozgunozerk: We don't have a config for backend at the moment. I will try to move this function to frontend, thanks for the suggesiton

nazar-pc (approved):

PR #209, opened 2022-05-14
provide custom types

isSerge:


nazar-pc (approved):
Makes sense, but does `apiTypes` need to be re-exported now?

nazar-pc (approved):

PR #213, opened 2022-05-18
Conventional log folders

ozgunozerk:
Very trivial PR. 

Still, I put some comments below to put light onto changes and their reasonings

### DIFF
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -34,6 +34,7 @@ subspace-runtime = { git = "https://github.com/subspace/subspace", rev = "72b2c4
 subspace-service = { git = "https://github.com/subspace/subspace", rev = "72b2c44971a2db1266995a8c0fe1223ef09be784" }
 subspace-solving = { git = "https://github.com/subspace/subspace", rev = "72b2c44971a2db1266995a8c0fe1223ef09be784" }
 tokio = { version = "1.11.0", features = ["macros", "rt-multi-thread"] }
+dirs-next = "2.0.0"
 
 [target.'cfg(windows)'.dependencies]
 winreg = "0.10.1"
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -13,6 +13,7 @@ tauri-build = { version = "1.0.0-rc.7", features = [] }
 
 [dependencies]
 anyhow = "1.0.44"
+dirs = "4.0.0"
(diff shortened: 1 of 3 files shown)


src-tauri/src/main.rs
```diff
@@ -90,8 +90,15 @@ fn get_this_binary() -> PathBuf {
 
 #[tokio::main]
 async fn main() -> Result<()> {
+    let ctx = tauri::generate_context!();
+    let id = &ctx.config().tauri.bundle.identifier;
```
ozgunozerk, line 94: we need the bundler name, and only way is to use `generate_context`.

src-tauri/src/main.rs
```diff
@@ -145,7 +152,7 @@ async fn main() -> Result<()> {
                 frontend_info_logger
             ],
         )
-        .build(tauri::generate_context!())
+        .build(ctx)
         .expect("error while running tauri application");
```
ozgunozerk, line 156: that's why, we replaced this one with the previously generated context :)
I tested this locally, both in DEV and PROD, it worked.

src-tauri/Cargo.toml
```diff
@@ -34,6 +34,7 @@ subspace-runtime = { git = "https://github.com/subspace/subspace", rev = "72b2c4
 subspace-service = { git = "https://github.com/subspace/subspace", rev = "72b2c44971a2db1266995a8c0fe1223ef09be784" }
 subspace-solving = { git = "https://github.com/subspace/subspace", rev = "72b2c44971a2db1266995a8c0fe1223ef09be784" }
 tokio = { version = "1.11.0", features = ["macros", "rt-multi-thread"] }
+dirs-next = "2.0.0"
```
nazar-pc, line 37: Why `dirs-next` and not more popular `dirs` that we use in node and farmer?

nazar-pc (approved):

src-tauri/Cargo.toml
ozgunozerk: Did not know about that, will update accordingly, thanks!

nazar-pc (approved):

nazar-pc:
On Linux this stored logs in in `/home/nazar-pc/.local/share/com.subspace.desktop` and the rest of data in `/home/nazar-pc/.local/share/subspace-desktop`, why? Why can't we have `subspace-desktop` instead of `com.subspace.desktop`?

PR #214, opened 2022-05-18
Fix csp for wasm

ozgunozerk:
This fixes the issue of not being able to generate a keyring with `sr25519` type.

This PR also makes our CSP more strict. To my testing, making it stricter did not create any issue. 
I followed Fabian's suggestions and made a small change on it: https://github.com/tauri-apps/tauri/issues/4160#issuecomment-1130363459

The small change: `style-src 'unsafe-inline' 'self';` -> this means: for css, we are enabling inline scripts execution. 
Actually, our app works without this as well, but becomes more ugly. Here is the difference

with this change:
image

without this change:
image

It is not recommended to enable this, I don't know how risky it will be for CSS files. It is up for discussion

### DIFF
--- a/src/lib/client.ts
+++ b/src/lib/client.ts
@@ -184,12 +184,17 @@ export class Client {
   }
 
   public async createRewardAddress(): Promise<string> {
-    const mnemonic = mnemonicGenerate()
-    const keyring = new Keyring({ type: 'sr25519', ss58Format: 2254 }) // 2254 is the prefix for subspace-testnet
-    await cryptoWaitReady();
-    const pair = keyring.createFromUri(mnemonic)
-    this.mnemonic = mnemonic
-    return pair.address
+    try {
+      const mnemonic = mnemonicGenerate()
+      const keyring = new Keyring({ type: 'sr25519', ss58Format: 2254 }) // 2254 is the prefix for subspace-testnet
+      await cryptoWaitReady();
+      const pair = keyring.createFromUri(mnemonic)
+      this.mnemonic = mnemonic
+      return pair.address
+    } catch (error) {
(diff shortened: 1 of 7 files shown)


src/lib/client.ts
```diff
@@ -184,12 +184,17 @@ export class Client {
...
+    try {
+      const mnemonic = mnemonicGenerate()
+      const keyring = new Keyring({ type: 'sr25519', ss58Format: 2254 }) // 2254 is the prefix for subspace-testnet
+      await cryptoWaitReady();
+      const pair = keyring.createFromUri(mnemonic)
+      this.mnemonic = mnemonic
+      return pair.address
+    } catch (error) {
+      util.errorLogger(error)
+      return "???"
```
isSerge, line 196: do we actually display this in case of error?

ozgunozerk: Let me update this with `""` instead, that will make more sense, because there are checks for `rewardAddress !== ""`. Will be better for logging. Thanks

isSerge (approved):

ozgunozerk:
We probably don't need `unsafe-inline` for styling. The color styling is coming from `appData.ts` as:
```ts
    Dialog.create({
      title: `Confirm selected directory.`,
      message: `
    <p style="font-size:12px;">
      Seleted plot directory.</br>
    </p>
    <p style="font-size:14px; color: #2196f3">
      <b>${plotDirectory}</b>
    </p>
`,
```

Although we have `self` as the source for `style-src`, it seems `appData.ts` is not covered in the scope of `self`. I will research more on this and get rid of `unsafe-inline`.

isSerge (approved):

PR #219, opened 2022-05-20
Edit node name

isSerge:
- add input to edit node name
- add node name to global state (instead of emitting event)
- minor refactoring of the global state and client

### DIFF
--- a/src/boot/init.ts
+++ b/src/boot/init.ts
@@ -4,6 +4,7 @@ import { globalState } from "../lib/global"
 import { Client } from "../lib/client"
 import { createApi } from '../lib/util';
 import { AutoLauncher } from "../lib/autoLauncher"
+import { appConfig } from "../lib/appConfig";
 
 const LOCAL_RPC = process.env.LOCAL_API_WS || "ws://localhost:9944"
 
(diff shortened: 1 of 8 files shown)


src/boot/init.ts
```diff
@@ -15,10 +16,10 @@ declare module "@vue/runtime-core" {
 }
 
 export default boot(async ({ app }) => {
-  const initActions = [
-    globalState.init()
-  ]
-  await Promise.all(initActions)
+  await appConfig.init()
+  const { nodeName } = (await appConfig.read());
```
ozgunozerk, line 20: what if nodeName was not set in the config (first time start of the application)?

isSerge: it will use initial value - empty string

ozgunozerk (approved):

src/boot/init.ts
```diff
@@ -15,10 +16,10 @@ declare module "@vue/runtime-core" {
...
+  globalState.setNodeName(nodeName);
+  await globalState.loadLangData()
```
nazar-pc, line 22: I don't get this refactoring, old and new code seems to be unrelated and changes do not correspond to commit messages :shrug:

src/layouts/MainLayout.vue
```diff
@@ -37,37 +53,56 @@ import * as process from 'process'
...
-
   data() {
     return {
       lang,
       global: global.data,
       appVersion: "",
       util,
       autoLaunch: false,
-      nodeName: ''
+      isEditName: false,
```
nazar-pc, line 70:
```suggestion
      isEditingName: false,
```

nazar-pc (commented):
Not sure how things work here to review properly I'm afraid

src/boot/init.ts
isSerge: this part is related to the prev commit, where I have changed `Global` class

nazar-pc (approved):

PR #220, opened 2022-05-21
node connection fixes for production

ozgunozerk:
This PR fixes 3 things:
- it turns out, we are using the port `3000`, not `8080` for production. So I added that entry as well to the node config.
- also, in production, a connection must be established between node and frontend as usual, but CSP was preventing that, so added the ws endpoint of node to CSP
- changed bundle identifier name from `com.subspace.desktop` to `subspace-desktop`

This should be a hot fix, and I will release 0.5.1 after this merges


nazar-pc (dismissed):
On ports side, would be nice to change port `9944` on the node to something else, so that CLI doesn't overlap with desktop app.

ozgunozerk:
Good idea, let me do that real quick

nazar-pc:
BTW ports 8080 and 3000 are VERY common and will definitely cause collisions on machines of developers. Better use something less common.

src-tauri/src/node.rs
```diff
@@ -199,7 +199,7 @@ fn create_configuration<CS: ChainSpec + 'static>(
                 other: ExecutionStrategy::NativeElseWasm,
             },
             rpc_http: Some("127.0.0.1:9933".parse().expect("http endpoint is valid")),
```
nazar-pc, line 201: HTTP was introduced in previous PR, but I don't think it needs to be exposed at all

nazar-pc (approved):

src-tauri/src/node.rs
ozgunozerk: true, ok

nazar-pc (approved):

Issue #206, opened 2022-05-12
Unable to fetch peers


isSerge:
I am running app v0.5.0 in dev mode. While syncing it shows 0 peers on a dashboard:

<img width="385" alt="image" src="https://user-images.githubusercontent.com/13568875/168052020-1fe59a88-4d54-47c0-928b-e5498fa0f204.png">

Also it shows errors in the console:

<img width="1015" alt="image" src="https://user-images.githubusercontent.com/13568875/168052097-934b9ae4-5827-47b7-88e0-317e5529eb2d.png">

However `api.rpc.rpc.methods` returns an array of methods and `system_peers` is there. Tried connecting to both public RPC and local node

ozgunozerk:
We have an additional error after the latest pr #207 
<img width="809" alt="image" src="https://user-images.githubusercontent.com/32795992/168418353-5d0fb349-9af4-4a81-b90f-839d63944b1d.png">

isSerge:
@ozgunozerk this is same error as was originally reported, that PR was not fixing it

ozgunozerk:
I was not referring to `peers()` error. I was referring to `SubPreDigest` one

isSerge:
@ozgunozerk #209

isSerge:
Fixed by #222

PR #222, opened 2022-05-23
Allow unsafe RPC methods

isSerge:
Fix for #206


nazar-pc (commented):
Comment about why it is needed and when can be removed would be nice

nazar-pc (approved):

ozgunozerk (dismissed):

nazar-pc (approved):

PR #224, opened 2022-05-26
bug fix and remove no longer needed plot progress tracker

ozgunozerk:
Probably some copy-pasta error, we changed the plot size back to 1024*1024 from 1024*1024*1024. 
Which was causing 20GB plot to effectively turn into 20MB plot.

Also, now that we changed how plottingProgress screen works, we don't need `plot_progress_tracker` function from backend. We don't use that information anymore. So I removed them (we can always get them back from git history if we need to)


i1i1 (approved):
LGTM. Glad we found it :)

i1i1 (approved):

PR #225, opened 2022-05-26
disable edit name

isSerge:
Temporarily disable edit node name until node restart is implemented on the backend


ozgunozerk (approved):

nazar-pc:
You can just add a notification for user to restart the app for changes to take place

isSerge:
@nazar-pc that is an option too, we can discuss it at the meeting

ozgunozerk:
@nazar-pc I strongly disagree on this one.
Users will not read it even though we put it in the notes, will think that it is not working, and we will get constant questions and bug reports about a thing that we already know `is not working`. 
When we have node restart functionality in the code, we can re-enable this feature. It is not a big feature as well, and works as is right now.

nazar-pc:
Well, restarting to change some settings is a standard practice in so many apps that I don't see it as a big problem.

PR #226, opened 2022-05-26
plot card render bug fixed

ozgunozerk:
thanks typescript

### DIFF
--- a/src/lib/appConfig.ts
+++ b/src/lib/appConfig.ts
@@ -1,5 +1,5 @@
 import { Dialog, DialogChainObject } from "quasar"
-import { appName, errorLogger } from "./util"
+import { appName, errorLogger, toFixed } from "./util"
 import * as path from "@tauri-apps/api/path"
 import * as fs from "@tauri-apps/api/fs"
 
@@ -67,6 +67,7 @@ export const appConfig = {
   async read(): Promise<Config> {
     const result = await fs.readTextFile(await this.configFullPath())
     const config: Config = JSON.parse(result)
+    config.plot.sizeGB = toFixed(config.plot.sizeGB, 2)
     return config
   },
   async write(config: Config): Promise<void> {
--- a/src/lib/appConfig.ts
+++ b/src/lib/appConfig.ts
@@ -67,7 +67,7 @@ export const appConfig = {
   async read(): Promise<Config> {
     const result = await fs.readTextFile(await this.configFullPath())
     const config: Config = JSON.parse(result)
-    config.plot.sizeGB = toFixed(config.plot.sizeGB, 2)
+    config.plot.sizeGB = toFixed(Number(config.plot.sizeGB), 2)
     return config
   },
   async write(config: Config): Promise<void> {
--- a/src/lib/appConfig.ts
+++ b/src/lib/appConfig.ts
@@ -67,6 +67,7 @@ export const appConfig = {
   async read(): Promise<Config> {
     const result = await fs.readTextFile(await this.configFullPath())
     const config: Config = JSON.parse(result)
+    // TODO: there maybe a better solution, or `sizeGB` should be string in the first place
     config.plot.sizeGB = toFixed(Number(config.plot.sizeGB), 2)
     return config
   },
--- a/src/lib/util.ts
+++ b/src/lib/util.ts
@@ -50,7 +50,8 @@ export async function infoLogger(info: unknown): Promise<void> {
 
 export function toFixed(num: number, precision: number): number {
   if (!num) return 0
-  return parseFloat(num.toFixed(precision))
+  // don't remove `Number` cast, it is required. Thanks TypeScript!
+  return parseFloat(Number(num).toFixed(precision))
 }
 
 export function plotTimeMsEstimate(gb: number): number {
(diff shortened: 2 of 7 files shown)


src/lib/util.ts
```diff
@@ -50,7 +50,8 @@ export async function infoLogger(info: unknown): Promise<void> {
 
 export function toFixed(num: number, precision: number): number {
   if (!num) return 0
-  return parseFloat(num.toFixed(precision))
+  // don't remove `Number` cast, it is required. Thanks TypeScript!
```
nazar-pc, line 53: Why is number cast necessary if input is supposed to be number already? Looks like a bug elsewhere to me.

ozgunozerk: Exactly. when I print the type of `num`, it was string??? 
Isn't typescript forcing it to be `number`?...

nazar-pc: There are a few strictness checks in TypeScript, I think not all of them might be enabled or there is `as number` or something like that somewhere that compiler will blindly believe even if it is not true (just like with `unsafe` in Rust).

So better find where this function is called from and fix that instead of hacking here.

src/lib/appConfig.ts
```diff
@@ -67,6 +67,7 @@ export const appConfig = {
   async read(): Promise<Config> {
     const result = await fs.readTextFile(await this.configFullPath())
     const config: Config = JSON.parse(result)
+    config.plot.sizeGB = toFixed(config.plot.sizeGB, 2)
```
nazar-pc, line 70: Not sure what this does, but it still seems wrong too.

```typescript
parseFloat(Number(num).toFixed(precision))
```

that you have in `toFixed` can still lead to numbers with a lot of digits after `.`, there is a reason why `Number.toFixed()` returns a string and not another float.

`toFixed` should return a string and be used directly in the layout template, not parsed back to float and reassigned into the config.

ozgunozerk: how come toFixed can still lead to a number with lot of digits after `.` ? We have given a precision of 2, not something huge. I think the language can handle 2 digits for the precision and not return an unexpected case?

However, the changed line should have been `toFixed(Number(config.plot.sizeGB), 2)` fixing now.

ozgunozerk: We are kind of in a hurry to get the release of 0.5.2, so I say, let's merge this change for now. And keep the discussion alive. Because refactoring it will take time, `toFixed` is used in many places, and we need to do more testing in other places if we want to change the signature of the function). 
If we decide on refactoring it, we can always do so in the future. What do you say @nazar-pc ?

nazar-pc: Floats are complicated in computing, for instance:
```
0.1+0.7
0.7999999999999999
```

So it would be preferrable to store integers whenever possible. Ideally in the config too and only converting to floats for UI purposes later.

Why does config store gigabytes as a string though? It seems like an odd design decision and the root of the problem.

ozgunozerk: Leo designed it back then...

src/lib/appConfig.ts
```diff
@@ -67,6 +67,7 @@ export const appConfig = {
   async read(): Promise<Config> {
     const result = await fs.readTextFile(await this.configFullPath())
     const config: Config = JSON.parse(result)
+    config.plot.sizeGB = toFixed(Number(config.plot.sizeGB), 2)
```
nazar-pc, line 70: `Plot.sizeGB` has a type of `number`. If it is not a number in the first place (else why would you have to cast it to a number, right?), then something is broken somewhere else.

Please add a TODO here and we can merge this workaround. Things can be really bad if you lie about types in TypeScript. In some sense it is even worse than untyped JavaScript.

nazar-pc: If the config was already written in a broken way we'll have to keep this workaround for some time, but what is worse is that I don't think the writing was fixed, so we keep creating incorrect configs even if it is worked around here.

I strongly suggest to use [zod](https://www.npmjs.com/package/zod), I think we already have it in relayer, cc @isSerge

ozgunozerk: JSON.parse(result) is returning a string. It is reading a file in JSON format. But the inside config, there is an integer type. 
Casting `Number` on it, is in fact, a solution to the very root of the problem I guess. Is there a way to `JSON.parse` understand that the field should be integer?

Adding TODO's now though, thanks!

isSerge: Yes, we use zod. I can tackle this issue

nazar-pc (approved):

ozgunozerk:
updated version, now we are ready for release

src/lib/appConfig.ts
nazar-pc: > JSON.parse(result) is returning a string

No, it takes a string and returns an object. If string was `{"plot":{"sizeGB":"123"}}` then `config.plot.sizeGB` will be decoded as a string. However if string is ``{"plot":{"sizeGB":123}}` then it will be decoded as a number, thankfully JSON can distinguish between numbers and string, though it doesn't have notion of specific numeric types.

nazar-pc (approved):

PR #230, opened 2022-05-31
node-restart mechanism

ozgunozerk:
It is draft, because we also have to introduce front-end changes. Before that, to eliminate redundant work, I wanted to get your recommendations on whether this approach makes sense to you @nazar-pc and @i1i1.

**Important:** This only includes a restart mechanism for node, not for farmer. I tried the same approach with farmer but could not succeed. Reasons are below. As a workaround, we were already listening for errors on the farmer, and if it crashes (and it will crash during a node restart), we were restarting the farmer. I just added some timeout between the restarts of farmer, giving some time to node to fully restart itself and start listening to farmer.

Reasons why I failed implementing a restart mechanism for farmer is:
- first problem:
    - we are waiting on the handle of the farmer job (which is the `MultiFarming` struct)
    - i also need to clone that handle, so if a `restart` request comes, i should drop this handle
    - this handle (`MultiFarming` struct) does not implement clone/copy, so I cannot have the same handle in two places. 
- second problem:
    - because of the already existing restart mechanism we have, the code becomes too ugly when I tried to implement the logic
    - so, even if we solve the first problem, I still argue against having a `stop` mechanism for farmer. I think our restart mechanism will do fine.

### DIFF
--- a/src-tauri/src/farmer.rs
+++ b/src-tauri/src/farmer.rs
@@ -6,6 +6,7 @@ use subspace_core_primitives::PublicKey;
 use subspace_farmer::multi_farming::{MultiFarming, Options as MultiFarmingOptions};
 use subspace_farmer::{Identity, NodeRpcClient, ObjectMappings, Plot, RpcClient};
 use tokio::sync::mpsc::{channel, Receiver, Sender};
+use tokio::time::{sleep, timeout, Duration};
 
 #[tauri::command]
 pub(crate) async fn farming(path: String, reward_address: String, plot_size: u64) -> bool {
--- a/src-tauri/src/farmer.rs
+++ b/src-tauri/src/farmer.rs
@@ -4,7 +4,7 @@ use std::path::PathBuf;
 use subspace_core_primitives::PublicKey;
 use subspace_farmer::multi_farming::{MultiFarming, Options as MultiFarmingOptions};
 use subspace_farmer::{Identity, NodeRpcClient, ObjectMappings, Plot, RpcClient};
-use tokio::net::tcp::TcpStream;
+use tokio::net::TcpStream;
 use tokio::sync::mpsc::{channel, Receiver, Sender};
 use tokio::time::{sleep, timeout, Duration};
 
@@ -64,7 +64,7 @@ async fn farm(
 ) -> Result<(), anyhow::Error> {
     if let Err(error) = timeout(Duration::from_secs(20), async {
         loop {
-            if TcpStream::connect("127.0.0.1:9947").is_ok() {
+            if TcpStream::connect("127.0.0.1:9947").await.is_ok() {
                 break;
             } else {
                 sleep(Duration::from_millis(500)).await;
--- a/src-tauri/src/farmer.rs
+++ b/src-tauri/src/farmer.rs
@@ -1,5 +1,6 @@
 use anyhow::{anyhow, Result};
-use log::info;
+use log::{error, info};
+use std::net::TcpStream;
 use std::path::PathBuf;
 use subspace_core_primitives::PublicKey;
 use subspace_farmer::multi_farming::{MultiFarming, Options as MultiFarmingOptions};
@@ -60,7 +61,22 @@ async fn farm(
     plot_size: u64,
     error_sender: Sender<()>,
 ) -> Result<(), anyhow::Error> {
+    if let Err(error) = timeout(Duration::from_secs(20), async {
+        loop {
+            if TcpStream::connect("127.0.0.1:9947").is_ok() {
+                break;
+            } else {
+                sleep(Duration::from_millis(500)).await;
+            }
+        }
+    })
+    .await
+    {
+        error!("Node is not responding for 20 seconds, farmer is unable to start");
+        return Err(anyhow!(error));
+    }
     raise_fd_limit();
+
     let reward_address = if let Some(reward_address) = reward_address {
         reward_address
     } else {
--- a/src-tauri/src/farmer.rs
+++ b/src-tauri/src/farmer.rs
@@ -1,10 +1,10 @@
 use anyhow::{anyhow, Result};
 use log::{error, info};
-use std::net::TcpStream;
 use std::path::PathBuf;
 use subspace_core_primitives::PublicKey;
 use subspace_farmer::multi_farming::{MultiFarming, Options as MultiFarmingOptions};
 use subspace_farmer::{Identity, NodeRpcClient, ObjectMappings, Plot, RpcClient};
+use tokio::net::tcp::TcpStream;
 use tokio::sync::mpsc::{channel, Receiver, Sender};
 use tokio::time::{sleep, timeout, Duration};
 
--- a/src-tauri/src/node.rs
+++ b/src-tauri/src/node.rs
@@ -249,5 +249,3 @@ fn create_configuration<CS: ChainSpec + 'static>(
         force_new_slot_notifications: false,
     })
 }
-
-
--- a/src-tauri/src/node.rs
+++ b/src-tauri/src/node.rs
@@ -19,7 +19,7 @@ use std::path::PathBuf;
 use std::sync::Once;
 use subspace_runtime::GenesisConfig as ConsensusGenesisConfig;
 use subspace_service::{FullClient, NewFull, SubspaceConfiguration};
-use tokio::runtime::Handle;
+use tokio::{runtime::Handle, sync::Mutex, task::JoinHandle};
 
 static INITIALIZE_SUBSTRATE: Once = Once::new();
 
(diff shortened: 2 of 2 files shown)


src-tauri/src/node.rs
```diff
@@ -258,3 +260,19 @@ fn database_config(base_path: &Path, cache_size: usize, role: &Role) -> Database
         cache_size,
     }
 }
+
+pub async fn node_controller(path: String, node_name: String, command: &str) {
+    static mut NODE_HANDLE: Option<JoinHandle<()>> = None;
+    match command {
```
nazar-pc, line 266: This is Rust, please don't use strings where enums should be used.

src-tauri/src/node.rs
```diff
@@ -258,3 +260,19 @@ fn database_config(base_path: &Path, cache_size: usize, role: &Role) -> Database
...
+        "start" => unsafe {
+            NODE_HANDLE = Some(init_node(path.into(), node_name).await.unwrap());
+        },
```
nazar-pc, line 269: This is definitely not safe, please use `Mutex` if you want to mutate something that could be mutated concurrently instead of resorting to `unsafe`.

src-tauri/src/node.rs
```diff
@@ -258,3 +260,19 @@ fn database_config(base_path: &Path, cache_size: usize, role: &Role) -> Database
...
+        "restart" => unsafe {
```
nazar-pc, line 270: Why not `stop`? So you could call `Stop` and then `Start` to get effective restart.

src-tauri/src/main.rs
```diff
@@ -60,6 +60,8 @@ async fn farming(path: String, reward_address: String, plot_size: u64) -> bool {
             // if there is an error, restart another farmer, and start listening again, in a loop
             loop {
                 let result = error_receiver.recv().await;
+                // probably node restarted, so wait 7 seconds before starting farmer
+                thread::sleep(Duration::from_millis(7000));
```
nazar-pc, line 64: There should be no sleep here. Since farming requires node, you should be able to listen to some notification or observe in some other way that the node node has started. Just doing arbitrary timeout will work sometimes and will fail other times

src-tauri/src/node.rs
```diff
@@ -258,3 +260,19 @@ fn database_config(base_path: &Path, cache_size: usize, role: &Role) -> Database
...
+            drop(NODE_HANDLE.take());
```
nazar-pc, line 271: Does this also work?:
```suggestion
            NODE_HANDLE.take();
```

Also I'm confident this will do nothing for node stopping, you need to call `.abort()` to make it actually stop.

nazar-pc (changes requested):

src-tauri/src/node.rs
ozgunozerk: I first tried with Mutex, but using mutex inside static was not allowed :')

src-tauri/src/main.rs
ozgunozerk: opsie, the same logic is present also when starting farmer in the usual workflow. We will need to address that well...

src-tauri/src/main.rs
```diff
@@ -60,6 +60,8 @@ async fn farming(path: String, reward_address: String, plot_size: u64) -> bool {
...
                 match result {
                     // we have received an error, let's restart the farmer
```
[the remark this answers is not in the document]
ozgunozerk: Why it won't restart? In case of error, it will send an error using the mpsc::channel (farmer.rs -> line 69). 
And it will restart a new farmer, in case the previous farmer is crashed.

src-tauri/src/main.rs
nazar-pc: Ideally figure out when RPC starts and subscribe to something that indicates that, in the worst case try to connect to the TCP port where you expect RPC to be available in a loop with a small timeout until it becomes ready.

src-tauri/src/main.rs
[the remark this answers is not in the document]
nazar-pc: Because there is `unwrap()` below, If you unwrap `Err` task will exit (because tokio handles panics for you and I think will print a message as well, but not more than that).

src-tauri/src/node.rs
nazar-pc: :ok_hand:

src-tauri/src/main.rs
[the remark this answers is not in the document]
ozgunozerk: Inside the farmer code, we have this:
```rust
tokio::spawn(async move {
        let result = multi_farming.wait().await;
        match result {
            Err(error) => {
                log::error!("{error}");
                error_sender.send(()).await.unwrap() // this send should always be successful
            }
            Ok(_) => unreachable!("wait function should not return Ok()"),
        }
    });
```
If there will be error, yes, unwrap'ing it won't do much good for the logging. That's why we have `log::error!` above in case it fails.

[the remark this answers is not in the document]
ozgunozerk: I still don't think this is the case. 
When node is stopped whilst farmer is working in CLI, farmer was giving a related error (`connection dropped` or something like that). 
When node stops, wouldn't farmer trigger the `Err(error) => {` branch of the `match`?

src-tauri/src/main.rs
ozgunozerk: how about the new version? I just pushed some new code about this

src-tauri/src/main.rs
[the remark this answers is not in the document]
ozgunozerk: Here is it: `2022-06-02T13:32:44.484948Z ERROR jsonrpsee_core::client::async_client: Error: Connection(Closed) terminating client`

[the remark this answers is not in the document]
nazar-pc: Hm, yes, you're right!

In that case I'm not sure you even need `match`, `if let Err(error) = multi_farming.wait().await` would be more compact.

src-tauri/src/node.rs
```diff
@@ -261,18 +266,15 @@ fn database_config(base_path: &Path, cache_size: usize, role: &Role) -> Database
...
     static mut NODE_HANDLE: Option<JoinHandle<()>> = None;
     match command {
-        "start" => unsafe {
+        NodeCommands::Start => unsafe {
             NODE_HANDLE = Some(init_node(path.into(), node_name).await.unwrap());
         },
-        "restart" => unsafe {
-            drop(NODE_HANDLE.take());
+        NodeCommands::Restart => unsafe {
+            NODE_HANDLE.take().unwrap().abort();
```
nazar-pc: This will panic if frontend sends two `Restart` commands quickly in a row. You need `if let Some() = NODE_HANDLE.take()` here.

src-tauri/src/farmer.rs
```diff
@@ -15,7 +17,23 @@ pub(crate) async fn farm(
     plot_size: u64,
     error_sender: Sender<()>,
 ) -> Result<(), anyhow::Error> {
+    let mut not_connected = true;
+    if let Err(error) = timeout(Duration::from_secs(20), async {
+        while not_connected {
+            if let Ok(_) = TcpStream::connect("127.0.0.1:9947") {
+                not_connected = false;
+            } else {
+                sleep(Duration::from_millis(1000)).await;
```
nazar-pc, line 25: I'd make this smaller, like 500 or 100, but yes, this is what I meant.

nazar-pc (commented):
Mostly looks good. Just minor comments and merge conflicts to resolve and we can land this.

src-tauri/src/node.rs
ozgunozerk: That possibility slipped from my attention, thanks!

ozgunozerk:
I got rid of the `restart` logic, because I discovered it is possible to handle both functionality (start and restart) under the `node_controller` function, so decided against more complicated logic and longer code.

Now: `node_controller` checks for if the there is a handle. If there are no handles, it starts a new node instance.
If there is a handle already present, if takes it, aborts the current instance, and then, starts a new node instance. 
Hence, no need to burden extra logic on the caller functions and main.

src-tauri/src/farmer.rs
```diff
@@ -60,7 +61,22 @@ async fn farm(
     plot_size: u64,
     error_sender: Sender<()>,
 ) -> Result<(), anyhow::Error> {
+    if let Err(error) = timeout(Duration::from_secs(20), async {
+        loop {
+            if TcpStream::connect("127.0.0.1:9947").is_ok() {
```
i1i1, line 67: Just a minor thing, but it would be better to use `tokio::net::TcpStream` here, as `std` one blocks future

ozgunozerk: thanks!

i1i1 (approved):

i1i1 (dismissed):

src-tauri/src/node.rs
```diff
@@ -51,10 +51,10 @@ impl NativeExecutionDispatch for ExecutorDispatch {
 
 #[tauri::command]
 pub(crate) async fn start_node(path: String, node_name: String) {
-    init_node(path.into(), node_name).await.unwrap();
+    node_controller(path, node_name).await;
```
nazar-pc, line 54: I'd inline that function, especially with `node_controller` name

nazar-pc (dismissed):

nazar-pc (approved):

i1i1 (approved):

nazar-pc (approved):

i1i1 (approved):

PR #231, opened 2022-05-31
Start Devtools programmatically in dev

isSerge:
Improves dev experience


ozgunozerk (commented):
why not like this:
```
app.get_window("main").unwrap().open_devtools();
```
and why are we immediately closing the devtools after opening them?

isSerge:
@ozgunozerk it did not close actually, so I forgot to remove that part. Thanks

ozgunozerk:
heard some issues about this:
https://github.com/tauri-apps/tauri/issues/4170

maybe relevant, we better check if it is working before merging it

PR #234, opened 2022-06-01
Monorepo dependency update

ozgunozerk:
This PR consists of:
- updating monorepo and substrate dependencies
- further refactoring main, better structured code
- `tracing` instead of `logging`
- fix a small issue, where we display old farmed blocks even if they are farmed under a different reward address (now only shows the ones farmed under the current reward address)

Don't go commit by commit, reviewing altogether is fine.

### DIFF
--- a/src-tauri/src/node.rs
+++ b/src-tauri/src/node.rs
@@ -223,10 +223,9 @@ fn create_configuration<CS: ChainSpec + 'static>(
             telemetry_endpoints,
             default_heap_pages: None,
             offchain_worker: OffchainWorkerConfig::default(),
-            force_authoring: true,
-            // force_authoring: env::var("FORCE_AUTHORING")
-            //     .map(|force_authoring| force_authoring.as_str() == "1")
-            //     .unwrap_or_default(),
+            force_authoring: env::var("FORCE_AUTHORING")
+                .map(|force_authoring| force_authoring.as_str() == "1")
+                .unwrap_or_default(),
             disable_grandpa: false,
             dev_key_seed: None,
             tracing_targets: None,
--- a/src-tauri/chain-spec.json
+++ b/src-tauri/chain-spec.json
@@ -5,18 +5,18 @@
     "Custom": "Subspace Gemini 1"
   },
   "bootNodes": [
-    "/dns/bootstrap-0.gemini-1a.subspace.network/tcp/30333/p2p/12D3KooWF9CgB8bDvWCvzPPZrWG3awjhS7gPFu7MzNPkF9F9xWwc",
-    "/dns/bootstrap-1.gemini-1a.subspace.network/tcp/30333/p2p/12D3KooWLrpSArNaZ3Hvs4mABwYGDY1Rf2bqiNTqUzLm7koxedQQ",
-    "/dns/bootstrap-2.gemini-1a.subspace.network/tcp/30333/p2p/12D3KooWNN5uuzPtDNtWoLU28ZDCQP7HTdRjyWbNYo5EA6fZDAMD",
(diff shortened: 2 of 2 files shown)


ozgunozerk:
Will merge when we update the chain spec (new one with new bootnodes)

src-tauri/Cargo.toml
```diff
@@ -1,39 +1,59 @@
...
+version = "0.5.3"
 description = "Subspace desktop"
 authors = ["Subspace Labs <https://subspace.network>"]
 license = "Apache-2.0"
 repository = "https://github.com/subspace/subspace-desktop"
 edition = "2021"
 
 [build-dependencies]
-substrate-build-script-utils = { version = "3.0.0", git = "https://github.com/paritytech/substrate", rev = "5d3e7c4ee9f5c8e370022d5f3fa9723185710831" }
+substrate-build-script-utils = { version = "3.0.0", git = "https://github.com/paritytech/substrate", rev = "24bea4c3cba7479e2cf2976a21e8111dfda6b1cd" }
```
nazar-pc, line 11: We switched to our fork in https://github.com/subspace/subspace/pull/533 (same commit though) to make builds faster and use less disk space

src-tauri/src/node.rs
```diff
@@ -181,7 +186,9 @@ fn create_configuration<CS: ChainSpec + 'static>(
             // TODO: Change to constrained eventually (need DSN for this)
             state_pruning: Some(PruningMode::keep_blocks(1024)),
             keep_blocks: KeepBlocks::Some(1024),
-            wasm_method: WasmExecutionMethod::Compiled,
+            wasm_method: WasmExecutionMethod::Compiled {
+                instantiation_strategy: WasmtimeInstantiationStrategy::PoolingCopyOnWrite,
+            },
             wasm_runtime_overrides: None,
             #[cfg(not(target_os = "windows"))]
```
nazar-pc, line 193: This time Windows should be the same as other platforms, performance issue was fixed

src/components/farmedList.vue
```diff
@@ -84,7 +84,10 @@ export default defineComponent({
   },
   computed: {
     farmedBlocksList(): FarmedBlock[] {
-      return this.$client?.data?.farming.farmed || []
+      const result = this.$client.data?.farming.farmed.filter((block) => {
+        return block.rewardAddr == this.rewardAddress
```
nazar-pc, line 88: @isSerge looks like this update doesn't know anything about votes, right? That is where the most of rewards will come from though. It is an unsigned extrinsic, see https://github.com/subspace/subspace/pull/485

src-tauri/src/utils.rs
```diff
@@ -0,0 +1,85 @@
+use log::{debug, error, info};
+use serde::Serialize;
+use std::io::Write;
+use std::path::PathBuf;
+use tauri::{
+    api::{self},
```
nazar-pc, line 6: :shrug: 
```suggestion
    api,
```

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "subspace-desktop",
-  "version": "0.5.2",
+  "version": "0.5.3",
```
nazar-pc, line 3: Shouldn't it be `0.6.0` though?

src/components/farmedList.vue
isSerge: yes, we will have a discussion on how do we display rewards today

src-tauri/Cargo.toml
ozgunozerk: this is wonderful 🥳

nazar-pc (approved):

PR #243, opened 2022-06-04
fix version

ImmaZoni:
missed Subspace Desktop version in cargo.toml


nazar-pc:
Already fixed in `main`

PR #247, opened 2022-06-06
Reset and log extraction

ozgunozerk:
Code-wise -> trivial. 
But it will be wordy to explain the changes.

- 1st change: `reset` button re-added. But changed the wording significantly. Please read the new messages. Any recommendation is welcome
- 2nd change: another button for viewing the logs.
    - viewing, not exporting, since I did not want to make the UI cumbersome and pop up many dialogs for where to save the log file, getting permissions and etc. So, as a workaround, when you click `view logs`, it will open the file location of the log file, using the native file explorer for the OS.
    - to be able to view the log file, we needed the path to it. I'm not hardcoding the path, but getting it from tauri. 
    - we are using tauri's `shell` api to be able to use the native file explorer. And that needs a weird check in the `tauri.conf.json`. It needs a regex that should match against the provided path. Or else, it does not run the command. That's why I supplied regex of including the string of `subspace-desktop`.
- 3rd change: `private key` names are changed with `seed phrase`, since we decided on following a terminology, and we should be consistent. 

That's all, thanks for reading :D

### DIFF
--- a/src/components/SaveKeys.vue
+++ b/src/components/SaveKeys.vue
@@ -34,7 +34,7 @@
               style="max-width: 200px"
             )
     .row
-      p Seed phrases are your password for your subspace farmer and wallet, this cannot be changed, guessed(easily), or reset if lost. It is imperative that this is stored in a secure, safe location. Without the seed phrase, you will not have access to your funds. Furthermore, anyone who steals your seed phrase will be able to do as they please with your funds.
+      p Seed phrase is your password for your subspace farmer and wallet, this cannot be changed, guessed, or reset if lost. It is imperative that this is stored in a secure, safe location. Without the seed phrase, you will not have access to your funds. Furthermore, anyone who steals your seed phrase will be able to do as they please with your funds.
   .row.q-pt-md
     q-checkbox(:label="lang.userConfirm" size="lg" v-model="userConfirm" :disable="!revealKey")
 </template>
--- a/src/components/SaveKeys.vue
+++ b/src/components/SaveKeys.vue
@@ -3,7 +3,7 @@
   .row.justify-center.q-gutter-lg
     .col-auto
       q-icon(color="blue-3" name="vpn_key" size="80px")
-      p.q-ml-sm {{ lang.privateKey }}
+      p.q-ml-sm {{ lang.seedPhrase }}
     .col
       .row.justify-center
         .col-8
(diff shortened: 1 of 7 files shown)


src/loc/en.json
```diff
@@ -81,9 +81,12 @@
...
     "IncentivizedLabel": "Incentivized Testnet",
     "IncentivizedTooltip": "Rewards (testSSC) you earn during incentivized testnet, will contribute to your mainnet rewards!",
-    "reset": "Reset",
+    "reset": "Reset Node!",
     "autoStart": "Start on Boot",
     "willAutoLaunch": "Subspace will launch automatically during boot",
-    "willNotAutoLaunch": "Subspace will not launch automatically during boot"
+    "willNotAutoLaunch": "Subspace will not launch automatically during boot",
+    "reset_heading": "Are you sure you want to reset and start from scratch?",
+    "reset_explanation": " This will erase everything, and you will have to start from the beginning. Only do this if there is something wrong in the network, or you imported a wrong reward address. You will need to re-import your existing private key.",
```
ozgunozerk, line 89: this is the further explanation shown in the very same confirm dialog

ozgunozerk:
image

ImmaZoni (approved):
All Looks great, though it does appear in your screenshot that version is not updated 0.6.4 very minor, everything else looks great!

ozgunozerk:
Yeah, if we are not in a hurry, I usually like improving the version as a separate pr, which becomes handy (separators) when writing release notes.

src/components/SaveKeys.vue
```diff
@@ -34,7 +34,7 @@
               style="max-width: 200px"
             )
     .row
-      p Private keys are your password for your subspace farmer and wallet, this cannot be changed, guessed(easily), or reset if lost. It is imperative that this is stored in a secure, safe location. Without the Private Key you will not have access to your funds. Furthermore, anyone who steals your private keys will be able to do as they please with your funds.
+      p Seed phrases are your password for your subspace farmer and wallet, this cannot be changed, guessed(easily), or reset if lost. It is imperative that this is stored in a secure, safe location. Without the seed phrase, you will not have access to your funds. Furthermore, anyone who steals your seed phrase will be able to do as they please with your funds.
```
nazar-pc, line 37:
```suggestion
      p Seed phrase is your password for your subspace farmer and wallet, this cannot be changed, guessed, or reset if lost. It is imperative that this is stored in a secure, safe location. Without the seed phrase, you will not have access to your funds. Furthermore, anyone who steals your seed phrase will be able to do as they please with your funds.
```

nazar-pc (dismissed):

nazar-pc (approved):

PR #254, opened 2022-06-10
update dependencies

ozgunozerk:
Please check the below comment, I'm not sure of it

### DIFF
--- a/src-tauri/src/node.rs
+++ b/src-tauri/src/node.rs
@@ -1,7 +1,8 @@
-use anyhow::Result;
+use anyhow::{anyhow, Result};
 use cirrus_runtime::GenesisConfig as ExecutionGenesisConfig;
 use log::{error, warn};
 use sc_chain_spec::ChainSpec;
+use sc_client_api::HeaderBackend;
 use sc_executor::{NativeExecutionDispatch, WasmExecutionMethod, WasmtimeInstantiationStrategy};
 use sc_network::config::{NodeKeyConfig, Secret};
 use sc_service::config::{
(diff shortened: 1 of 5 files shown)


nazar-pc:
We need to include https://github.com/subspace/subspace/pull/581 or else it will not work.
Also take a look at https://github.com/subspace/subspace/pull/574 and replicate node key changes. Basically when we know that node has stuck at a particular block height we need to reset network keys because there is a high probability that node is banned by the rest of the network and will not be able to reach to anyone to sync.

src-tauri/src/node.rs
```diff
@@ -119,8 +120,35 @@ async fn create_full_client<CS: ChainSpec + 'static>(
...
 
-    subspace_service::new_full::<subspace_runtime::RuntimeApi, ExecutorDispatch>(config, true)
-        .map_err(Into::into)
+    let config_dir = config
+        .base_path
+        .as_ref()
+        .map(|base_path| base_path.config_dir("subspace_gemini_1b"));
+
+    // subspace_service::new_full::<subspace_runtime::RuntimeApi, ExecutorDispatch>(config, true)
+    //     .map_err(Into::into)
```
nazar-pc, line 129: I think this can be removed

nazar-pc (approved):
Looks fine to me, were there any issues with it so it wasn't released yesterday?

ozgunozerk:
> Looks fine to me, were there any issues with it so it wasn't released yesterday?

I think because it was not approved till today

nazar-pc (approved):

PR #262, opened 2022-06-14
added support for external drives

ozgunozerk:
Long-awaited support for external drives is here :)

It was more trivial than I thought :D


nazar-pc (approved):

ozgunozerk:
This is implemented for all OS's, yet not tested for all OS's. See the checklist above. It is not released, so version 0.6.7 does not have this feature yet. It will be in the next release though :)

PR #273, opened 2022-06-25
openssl fix for ubuntu & version bump

### LINKED ISSUE #272: openssl issue on Linux side
Here is the link to the forum: https://forum.subspace.network/t/error-while-loading-shared-libraries-libssl-so-1-1-cannot-open-shared-object-file-no-such-file-or-directory/447/4

Quick summary: because of the `openssl` dependency not being packed into the application, some of the `openssl` libraries in our user's machine may create some inconsistencies due to their version being different. 

To fix it, we will embed the `openssl` library into the executable itself.

ozgunozerk:
Fixes #272 

@ImmaZoni and I were testing this. I tried adding the `openssl` dependency to the frontend, via `yarn add openssl`. However, it did not work. 
So we tried adding it to `backend` side via `cargo add openssl`, and set `vendored` as the feature. It worked. 

Don't know why it did not work when we included it as a frontend dependency though...

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "subspace-desktop",
-  "version": "0.6.9",
+  "version": "0.6.10",
   "private": true,
   "description": "Subspace desktop",
   "author": "Subspace Labs <https://subspace.network>",
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -8793,7 +8793,7 @@ dependencies = [
 
 [[package]]
 name = "subspace-desktop"
-version = "0.6.9"
+version = "0.6.10"
 dependencies = [
  "anyhow",
  "cirrus-runtime",
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -5034,16 +5034,28 @@ dependencies = [
 
 [[package]]
 name = "openssl"
-version = "0.10.38"
+version = "0.10.40"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0c7ae222234c30df141154f159066c5093ff73b63204dcda7121eb082fc56a95"
+checksum = "fb81a6430ac911acb25fe5ac8f1d2af1b4ea8a4fdfda0f1ee4292af2e2d8eb0e"
 dependencies = [
  "bitflags",
  "cfg-if",
  "foreign-types",
  "libc",
  "once_cell",
+ "openssl-macros",
  "openssl-sys",
+]
+
+[[package]]
+name = "openssl-macros"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b501e44f11665960c7e7fcf062c7d96a14ade4aa98116c004b2e37b5be7d736c"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
 ]
 
 [[package]]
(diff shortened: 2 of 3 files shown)


isSerge (approved):

i1i1 (approved):

nazar-pc:
What OS is that, @ImmaZoni? There is no information about that on the forum and no link to original discussion.

It is very unusual to not have OpenSSL installed on the machine, but I'm also wondering why is it necessary in the first place, looks like Tauri bug to me that needs to be reported. If it was needed, OpenSSL would have been in dependencies already.

I feel like including OpenSSL for no reason and without any actual explanation is not the right approach.

BTW, does this actually fix an issue for user?

ozgunozerk:
The executable was not working in Justin's Linux too. 
And I built the Linux executable for this code, sent it to Justin, it worked on his machine. 

OpenSSL is actually in the dependencies of tauri, we checked the `cargo tree` with @i1i1. But it was not the `vendored` version of the OpenSSL (which embeds the dependency into the executable), that's why we added the dependency with `vendored` feature enabled. 

Also worth mentioning, this `openSSL` bug is not specific to version 0.6.9 of the application. Justin was getting the same library error for every other version he tried (and he even tried 0.3.3, along with all 0.6.x). In summary, this is not a new thing, and it was surprising to hear this issue for the first time in the forum

nazar-pc:
Did you report this to Tauri? I'm not opposed to workaround if it is clearly marked as such and has an upstream ticket corresponding to it.

ozgunozerk:
I previously did not think of it as a bug from the Tauri. I will report it to them now, thanks!

i1i1:
@nazar-pc it is clearly a tauri bug I think, as for the deb packages it doesn't include openssl as a dependency. I thought we could at least temporary solve the issue by vendoring openssl.

I guess we should at least create an issue for tauri about that.

nazar-pc:
Yes, and there should be a TODO in Cargo.toml saying we need to remove OpenSSL, it is there just for feature unification and isn't used otherwise.

ozgunozerk:
Found this: 
https://github.com/tauri-apps/tauri/issues/1355#issuecomment-824012536

nazar-pc:
Have you tried changing `ubuntu-latest` to `ubuntu-20.04`? Using `latest` for release builds is a bad idea because it is a moving target. With 20.04 it might build and run just fine both on 20.04 and 22.04.

ozgunozerk:
I haven't tried, but Justin's Ubuntu was 22.04... 
So my vote is on it is not related to that from straight-forward logic

i1i1:
Tried with docker on `ubuntu:latest` and it still missed `libssl.so`

ozgunozerk:
Reported this to tauri team in their discord, will post here if they answer 👍

ozgunozerk:
Fabian answered to my question os `My point was, why openssl is used, and why does it not embed the library into the executable if it is needed? Maybe tauri dependencies should add the vendored feature to the openssl dependency?`:

> 0) attohttpc doesn't have that option (our reqwest alternative does)
1) bigger bundle size
2) it needs more build dependencies (c compiler, perl and make; only perl could be an issue, c compiler and make should be installed if you installed tauri's deps)
3) Your app's ssl lib doesn't get security updates (this is pretty huge for openssl)

It's only a problem now because ubuntu2204 switched to openssl v3 and openssl versions are not compatible (fun fact: v1 and v1.1 are not compatible with each other either). But because of this we need to look into it no matter the pros/cons.

i1i1:
Hmm, makes sense, but at least they can add it as a deb package dependency then. This is a clear bug, as an app won't start if you install the package.

ozgunozerk:
I got rid of the unused `http` stuff in tauri config, and openssl problem disappeared. Will open a new PR soon with the changes, closing this one for now

nazar-pc:
> Hmm, makes sense, but at least they can add it as a deb package dependency then. This is a clear bug, as an app won't start if you install the package.

They can't as OpenSSL 1.1.x is no longer present in Ubuntu 22.04 repos anymore :upside_down_face: 

I'd argue they need to switch to OpenSSL-free library for those things.

BTW, I have prepared Meson support for OpenSSL that doesn't require Make, Perl, but of course still needs C compiler: https://github.com/mesonbuild/wrapdb/tree/master/subprojects/packagefiles/openssl

PR #276, opened 2022-07-04
Tauri updater

isSerge:
- adds updater to Tauri config
- adds Tauri update keys to GitHub actions workflow

Currently, it will check for updates on the app launch. We still need to clarify UX for periodical checks and dropdown menu item for updates (check for update, install update, etc.)

### DIFF
--- a/src-tauri/tauri.conf.json
+++ b/src-tauri/tauri.conf.json
@@ -44,7 +44,12 @@
       }
     },
     "updater": {
-      "active": false
+      "active": true,
+      "endpoints": [
+          "https://releases.myapp.com/{{target}}/{{current_version}}"
+      ],
+      "dialog": true,
+      "pubkey": "YOUR_UPDATER_SIGNATURE_PUBKEY_HERE"
     },
     "allowlist": {
       "all": true,
--- a/src-tauri/tauri.conf.json
+++ b/src-tauri/tauri.conf.json
@@ -46,10 +46,10 @@
     "updater": {
       "active": true,
       "endpoints": [
-        "https://tauri-update-server.serge988.workers.dev/v1/{{target}}/{{arch}}/v0.6.8"
+        "https://tauri-update-server.serge988.workers.dev/v1/{{target}}/{{arch}}/{{current_version}}"
       ],
       "dialog": true,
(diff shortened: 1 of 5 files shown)


src-tauri/tauri.conf.json
```diff
@@ -44,7 +44,12 @@
       }
     },
     "updater": {
-      "active": false
+      "active": true,
+      "endpoints": [
+        "https://tauri-update-server.serge988.workers.dev/v1/{{target}}/{{arch}}/{{current_version}}"
```
nazar-pc, line 49: Should this URL be updated from `serge988` to something else?

ch00mba: Do we need different domain for this? As I understand its for demo purposes only

ch00mba (approved):
looks ok for demo

ozgunozerk:
We are going to abandon tauri actions in the future because it doesnt meet the requirements we have now. 
I'm working on writing a new workflow and testing it, hopefully, it will finish soon.
What will happen to the updater then?

nazar-pc:
BTW, what are those requirements? Can we just submit a PR with additional capabilities to the official action or maintain a fork in the worst case?

isSerge:
@ozgunozerk it should work as long as we have gh actions workflow

ozgunozerk:
Requirements are the following:
- tauri actions only let us build the application for only one configuration (there are currently no way to provide both `normal` and `opencl` version for the same release)
- tauri actions also do not let us rename the executables (we won't be able to rename them with `opencl` and with `os_name`)

When I contacted them, they suggested using a custom workflow instead of using tauri-actions. 
We talked with Serge, and there may be lots of things to debug in tauri-actions. So, I went with re-writing our workflow without tauri-actions. It's here -> https://github.com/ozgunozerk/subspace-desktop-workflow/blob/main/.github/workflows/publish.yml

> @ozgunozerk it should work as long as we have gh actions workflow

How to use these keys for updated without `tauri-actions`?

src-tauri/tauri.conf.json
isSerge: @nazar-pc I will use another subdomain

isSerge:
@ozgunozerk as far as I can see `TAURI_PRIVATE_KEY` and `TAURI_KEY_PASSWORD` are not used in the `tauri-action` itself. We just need to set these env variables when we create builds and bundles

nazar-pc:
> @ozgunozerk as far as I can see TAURI_PRIVATE_KEY and TAURI_KEY_PASSWORD are not used in the tauri-action itself. We just need to set these env variables when we create builds and bundles

This doesn't make sense to me, why would we specify them if they are not used anyway?

nazar-pc:
Makes sense now :+1:

nazar-pc (dismissed):

ch00mba (approved):
:+1:

PR #278, opened 2022-07-07
opencl support

ozgunozerk:
This PR adds opencl support to Desktop Application,
also configures release profile


src-tauri/Cargo.toml
```diff
@@ -66,19 +66,22 @@ features = [
...
 
 [features]
 default = [ "custom-protocol" ]
 custom-protocol = [ "tauri/custom-protocol" ]
+opencl = ["subspace-farmer/opencl"]
 
 [profile.dev]
 opt-level = 3
 
+[profile.release]
```
nazar-pc, line 80: Why `release` and not `production`? It might be still desirable to build `release` in some cases while not spending quite as much time as with these parameters.

src-tauri/Cargo.toml
```diff
@@ -66,19 +66,22 @@ features = [
...
 
 [dependencies.tauri]
-features = ["clipboard-all", "dialog-all", "fs-all", "global-shortcut-all", "gtk-tray", "notification-all", "os-all", "path-all", "process-exit", "process-relaunch", "shell-all", "system-tray", "window-all", "updater"]
+features = ["clipboard-all", "dialog-all", "fs-all", "global-shortcut-all", "gtk-tray", "notification-all", "os-all", "path-all", "process-exit", "process-relaunch", "shell-all", "system-tray", "updater", "window-all"]
 version = "1.0.0-rc.11"
 
 [features]
 default = [ "custom-protocol" ]
 custom-protocol = [ "tauri/custom-protocol" ]
+opencl = ["subspace-farmer/opencl"]
```
nazar-pc, line 75: Did you confirm necessary OpenCL dependencies are actually in the `*.deb` package dependencies if you enable this feature though?

I don't see any related changes, but maybe Tauri is smart enough to do that automatically. Also I don't see CI pipeline changes, so this feature doesn't have any effect yet anyway.

nazar-pc (approved):

nazar-pc (dismissed):

src-tauri/Cargo.toml
ozgunozerk: I did not confirm they are in the deb package :/ will try to check 
Updated CI pipeline 👍

.github/workflows/ci-tests.yml
```diff
@@ -45,11 +45,19 @@ jobs:
       - name: install Rust nightly
         uses: actions-rs/toolchain@v1
         with:
-          toolchain: nightly-2022-02-15
+          toolchain: nightly-2022-05-16
```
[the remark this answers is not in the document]
ozgunozerk: these were updated in the other pr, i thought they were here also... thanks!

src-tauri/Cargo.toml
ozgunozerk: opencl library is not in the deb package as you guessed... ill look for a way to include them in the deb

ozgunozerk: found the way to specify dependencies in tauri conf json, trying to figure out doing that conditionally for our case `#[cfg(feature = "opencl")]`

i1i1:
Sorry for force pushing. Wanted to squash last 4 commits with the same name. The most interesting part is the last two commits.

nazar-pc (dismissed):
Looks good to me, I wish there was a way to extend config instead of duplicating it though.

BTW, we can add `ocl-icd-opencl-dev` dependency regardless of whether OpenCL support was compiled in, does it mean we would be able to use the same config and specify feature in CLI instead?

src-tauri/tauri.conf.opencl.json
```diff
@@ -0,0 +1,113 @@
...
+        "icons/icon.ico"
+      ],
+      "resources": [],
+      "externalBin": [],
+      "copyright": "",
+      "category": "DeveloperTool",
+      "shortDescription": "",
+      "longDescription": "",
+      "deb": {
+        "depends": ["ocl-icd-opencl-dev"]
```
nazar-pc, line 32: Dev package shouldn't be used in production dependencies, `ocl-icd-libopencl1` should be used instead, that is where `OpenCL.so` shared library comes from.

nazar-pc (commented):
I don't see in docs if it actually reads `tauri.conf.json` if you specify a custom path to config, 
Also convention for files there is to end with `.conf.json`, and you put `opencl` in between those.

Also I see it already reads files like `tauri.linux.conf.json` and merges into default config, so we should be able to use that and don't do anything special for it to be built with OpenCL on Linux at least (on Windows question remains about OpenCL.dll).

i1i1:
> I don't see in docs if it actually reads tauri.conf.json if you specify a custom path to config,

Tauri devs answered that `--config` has the same semantics here (merges default config with second one.

> Also I see it already reads files like tauri.linux.conf.json and merges into default config, so we should be able to use that and don't do anything special for it to be built with OpenCL on Linux at least (on Windows question remains about OpenCL.dll).

Hmm, but we don't need to install opencl runtime if we disable opencl feature. So I just think it will be better to have just one more file for a feature.

nazar-pc (commented):
> Tauri devs answered that --config has the same semantics here (merges default config with second one.

Makes sense then, but it wasn't in Tauri docs :upside_down_face: 

> Hmm, but we don't need to install opencl runtime if we disable opencl feature. So I just think it will be better to have just one more file for a feature.

It is just one package on Ubuntu, but I'm fine with it as is since configs are merged, but please rename config to be consistent in naming with the rest of configs used in Tauri.

nazar-pc (approved):

i1i1:
I believe tests pass, but ci fails because runners don't have enough space for a build. Can we force merge it?

ozgunozerk:
CI problem is solved. 
I separated the clippy-fmt checks to another job. 
Now the CI workflow should be faster compared to the previous runs, so 2 birds with 1 stone. 

I inspected the changes you guys made while I was away. All looking great to me, thanks!
If you don't have any other things in mind, I can update the branch and merge this. So, I will focus on the publish-workflow PR as the next step.

nazar-pc (approved):

PR #279, opened 2022-07-07
workflow update

ozgunozerk:
Abandons tauri action and replaces it with more manual and configurable commands.

Here are some notes that may be helpful for the future (and while I'm away):
1. I could not succeed building a `.deb` image for `arm linux`. Cross-compile dependencies were creating a lot of problems. Nazar suggested we can skip it for now, since it is no big deal at the moment.
2. I have confirmed the executable/installer will be successfully built on every platform in the workflow without any issue. However, `yarn tauri build` command requires signing for windows, and I've been working in my own repo, and did not include the signing keys to my repo for security reasons. In short, I don't think there will be an issue, but if an issue arises, here are the potential reaons:
    - executable name / path may be incorrect
    - signing process may be erroneous
    - I've checked every other thing, so no need to look at other places for debugging
3. I know I did not include the latest `updater` changes for the workflow. @isSerge , you are more than welcome to commit to this for introducing these changes :)
4. Set draft to true for now, since we will have more testing with this for updater and windows

Todo: 
- [ ] change line 52 to this after the current publish trial action is finished: `if: matrix.build.suffix == 'macos-aarch64'`


isSerge:
Need to add Tauri updater and test on Windows - assigning it to myself.

ozgunozerk:
There are 2 tauri related impediments:
1. we cannot use `-Z build-std` -> look here for the issue I opened: https://github.com/tauri-apps/tauri/issues/4720
2. tauri appends `--release` to the command. So I couldn't successfully append `--profile=production` to our command. That's why, I reverted the `production` profile to release. If we want to build it faster locally, we can use debug, or we can manually comment out the compile optimizations.

nazar-pc:
I'll wait for other fixes before looking into runtime building issues. We had it like that for a long time and never had such issues neither in monorepo nor here in tests and otherwise. I see no direct reason why would it start happening here now, so will see if it is still reproducible later.

ozgunozerk:
> I'll wait for other fixes before looking into runtime building issues. We had it like that for a long time and never had such issues neither in monorepo nor here in tests and otherwise. I see no direct reason why would it start happening here now, so will see if it is still reproducible later.

What are you referring to by ***other fixes***?

nazar-pc:
Well, last commit here installs `protobuf`, which as we discovered is not actually necessary, rather TypeScript config needs to be fixed instead.

ozgunozerk:
@nazar-pc 
on top of all the fixes, i rebased the `workflow-experimenting` branch on main, so it would use up to date dependencies for subspace, rust and substrate. The error is still there -> https://github.com/subspace/subspace-desktop/runs/7535033385?check_suite_focus=true

nazar-pc:
I don't see why it would happen, asked community here: https://users.rust-lang.org/t/build-script-dependency-doesnt-work-properly/78987

PR #280, opened 2022-07-11
Add internalization

isSerge:
- Introduces `vue-i18n` to manage internalization (recommended approach by Vue and Quasar)
- Removes hardcoded test and replaces it with translations


nazar-pc (approved):
Makes sense to me, but next time would be nice to not have 3rd commit at all and squash first and last together. You can do that cleanly before submitting a PR.

PR #282, opened 2022-07-12
UI optimization for syncing progress highestBlock value

isSerge:
only update the highest block if the new value is bigger

### DIFF
--- a/src/lib/client.ts
+++ b/src/lib/client.ts
@@ -11,9 +11,10 @@ import { getStoredBlocks, storeBlocks } from "./blockStorage"
 import {
   emptyClientData,
   FarmedBlock,
-  SubPreDigest
+  SubPreDigest,
+  SyncState
 } from "../lib/types"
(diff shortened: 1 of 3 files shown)


nazar-pc (commented):
This is not a sufficient fix. If you connect to a node on non-canonical fork that has a higher block height, application will get stuck in syncing state forever. This needs to be smarter than that somehow.

isSerge:
@nazar-pc in that case we could notify the user that he/she is on the non-canonical fork and provide steps to fix this. But it seems like a separate task to me

nazar-pc:
That is the problem: we do not know. There was a higher block number observable, but we can't know if that was the canonical chain we should reach eventually or some random fork we should ignore.

What about remembering highest block number of some time, but revert back to the currently visible in case it doesn't change much for a long time? Otherwise I'd leave the behavior as is, at least it accurately shows what node is currently thinking the sync target is.

PR #283, opened 2022-07-18
Small mnemonic modal update

isSerge:
Pass mnemonic as a component prop only once instead of storing and removing it on unmount


ozgunozerk (approved):
lgtm

PR #295, opened 2022-07-27
reflect monorepo fix

ozgunozerk:
see: https://github.com/subspace/subspace/pull/721

### DIFF
--- a/src-tauri/src/farmer.rs
+++ b/src-tauri/src/farmer.rs
@@ -131,7 +131,7 @@ async fn farm(base_directory: PathBuf, farm_args: FarmingArgs) -> Result<(), any
 
     // Starting the relay server node.
     let (relay_server_node, mut relay_node_runner) = subspace_networking::create(Config {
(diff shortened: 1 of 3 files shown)


src-tauri/src/farmer.rs
```diff
@@ -169,7 +169,8 @@ async fn farm(base_directory: PathBuf, farm_args: FarmingArgs) -> Result<(), any
             enable_dsn_archiving: matches!(archiving, ArchivingFrom::Dsn),
             enable_dsn_sync: dsn_sync,
             enable_farming: true,
-            relay_server_node,
+            listen_on,
```
i1i1, line 172: I believe you just need to pass empty vector here. We don't care at which interface each plot will listen

ozgunozerk: will do

i1i1 (approved):

nazar-pc (approved):

PR #296, opened 2022-07-27
re-introduce node restart

### LINKED ISSUE #197: Persist node name across restarts

ozgunozerk:
This pr:
- introduces node restart feature, and hence fixes #197 
- also makes the `nodeRestart` function return a boolean for success and failure
- and accordingly, handles these errors on the frontend side as well

in case of any unexpected failure happens in restart process, both farmer and node will try repeatedly for 10 seconds, and after that quit trying. So, the application will never get stuck in a loop because of this restart mechanism


i1i1 (approved):

src-tauri/src/farmer.rs
```diff
@@ -113,6 +115,20 @@ async fn farm(base_directory: PathBuf, farm_args: FarmingArgs) -> Result<(), any
...
+            if NodeRpcClient::new(&node_rpc_url).await.is_ok() {
+                break;
+            } else {
+                sleep(Duration::from_millis(500)).await;
+            }
+        }
+    })
+    .await
+    {
+        error!("Node is not responding for 10 seconds, farmer is unable to start");
```
nazar-pc, line 129: Is it recoverable from this state though?

src-tauri/src/node.rs
```diff
@@ -52,11 +53,49 @@ impl NativeExecutionDispatch for ExecutorDispatch {
...
+
+    // if there is already a node running, stop it
+    if let Some(guard) = node_handle_guard.take() {
+        guard.abort();
+    }
+
+    // wait for the previous instance to finish if new instance is unable to start
+    // we cannot know the previous node instance is completely finished for now
+    // track the issue from here: https://github.com/paritytech/substrate/issues/11654 and https://github.com/subspace/subspace/issues/578
+    if let Err(_) = timeout(Duration::from_secs(10), async {
```
nazar-pc, line 72: This should be a fallback with TODO to remove in the future. I'd argue it is still helpful to have a loop with upgrade step too, it feels wrong to hit an error until you don't as the default behavior.

src-tauri/src/node.rs
```diff
@@ -52,11 +53,49 @@ impl NativeExecutionDispatch for ExecutorDispatch {
...
+            match init_node(path.clone().into(), node_name.clone()).await {
+                Ok(guard) => {
+                    *node_handle_guard = Some(guard);
+                    break;
+                }
+                Err(err) => {
+                    error!(
+                        "could not start the node with err: {err}, will wait for 100 milliseconds"
+                    );
+                    sleep(Duration::from_millis(200)).await;
```
nazar-pc, line 83: This isn't 100 despite message says so

src-tauri/src/farmer.rs
```diff
@@ -113,6 +115,20 @@ async fn farm(base_directory: PathBuf, farm_args: FarmingArgs) -> Result<(), any
...
+        return Err(anyhow!(error));
+    }
     let archiving_client = NodeRpcClient::new(&node_rpc_url).await?;
```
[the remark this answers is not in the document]
ozgunozerk: Thanks! The reason I did not choose this way is:
we have both `archiving_client` and `farming_client`. 
Both of them are to be used later in the code... Initiating one of them in this complex logic, and the other one in a single line separated from this logic, might look weird and confusing :/

src-tauri/src/farmer.rs
ozgunozerk: I don't think so. There should be something wrong with the node in this case. Probably a restart on the app would fix it

src-tauri/src/node.rs
```diff
@@ -57,35 +61,51 @@ impl NativeExecutionDispatch for ExecutorDispatch {
...
 
     // if there is already a node running, stop it
-    if let Some(guard) = node_handle_guard.take() {
-        guard.abort();
+    if let Some(handle) = node_handle_guard.take() {
+        handle.abort();
+        while !handle.is_finished() {
+            sleep(Duration::from_millis(200)).await;
+        }
+        let weak_client = weak_client_guard.take().unwrap(); // safe to unwrap, since we know a previous instance is present
```
nazar-pc, line 80: If safe then use `.expect()`?

ozgunozerk: Isn’t the only difference between `expect` and `unwrap` is, 
`expect` = `unwrap` + message ?

Why should I prefer `expect` over `unwrap` if safe?
If it was not safe, i shouldn't have `unwrapped` at all, and handled the error instead.

nazar-pc: `.expect()` forces you to provide an explicit message that convinces reviewer why it is okay to do that. We don't use `.unwrap()` at all in the node's code and we plan to eliminate it in farmer's code too. Substrate has the same policy.

ozgunozerk: now I understood :)

src-tauri/src/farmer.rs
```diff
@@ -206,11 +208,17 @@ async fn farm(base_directory: PathBuf, farm_args: FarmingArgs) -> Result<(), any
         match result {
             Err(error) => {
                 error!("{error}");
-                error_sender.send(()).await.unwrap() // this send should always be successful
+                error_sender
+                    .send(())
+                    .await
+                    .expect("error receiver is always listening")
```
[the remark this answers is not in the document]
ozgunozerk: I actually care about the `receiver` being alive. The `receiver` is in `farming` function, and it should always be alive. That's why I put `expect` here. 
I get that it is not obvious from the context of the function. But it should be clear after inspecting `farming` function. 

I will put a more descriptive expect message. If you still think it is not a good idea, I can use error handling in here, because I want to be sure that `receiver` is alive.

src-tauri/src/farmer.rs
```diff
@@ -208,17 +210,15 @@ async fn farm(base_directory: PathBuf, farm_args: FarmingArgs) -> Result<(), any
         match result {
             Err(error) => {
                 error!("{error}");
-                error_sender
-                    .send(())
-                    .await
-                    .expect("error receiver is always listening")
+                error_sender.send(()).await.expect(
+                    "the receiver is defined in the `farming` function, it never stops listening",
```
nazar-pc, line 214: I understand that something somewhere might actually be holding it, but as far as this function is concerned, it is possible, very easily in fact, to make this panic. Why don't you just print an error about it, why do you need `.expect()`?

Also I don't quite understand why background task is needed here. You could return `multi_plots_farm` from this function and then call `.wait()` on the higher level where both sender and receiver sides of the channel are created and you can actually guarantee it will not panic.

nazar-pc (approved):
Approving to move forward, but I'm not fully happy still

src-tauri/src/farmer.rs
ozgunozerk: That sounds more direct than my approach, I will try that :) Thanks!

ozgunozerk: Tried it, did not like it because of code duplicaiton. 
In the higher level one:
- we are starting `farm` instance (1),
- we start listening for errors on that `farm` instance
- if any error is sent, we are starting a new `farm` instance (2)

so, this `wait()` mechanism and the long code for sending error messages after that are duplicated. 
I can put these duplicated code pieces into a function, but then it is no different than before. 

And this code duplication is around 20 lines, making the higher level function unreadable.

----

The reason I wanted to utilize `expect` instead of printing an error about it is:
I think it is not possible for the `receiver` to stop listening, and if that occurs, something should have been really off in the higher level function. In this case, a new `farm` instance won't be created successfully, and the main functionality of the desktop app will be hindered. That's why, panicking seemed like a good choice.

nazar-pc: Something like this?:
```rust
loop {
    if let Err(error) = start_farm.await {
        // TODO
    }
}
```

Shouldn't have much code duplication, just a matter or restructuring it.

> I think it is not possible for the receiver to stop listening, and if that occurs, something should have been really off in the higher level function

My issue with this is that it is not possible to guarantee just within this function. It is ideal for functions to guarantee lack of panics statically.

PR #298, opened 2022-07-29
Add error page

isSerge:
image

- add error page
- specific errors are reflected in the error title
- body text is generic (until we clarify resolutions for specific errors)


nazar-pc (commented):
I have a few questions about UI:
1. Do I understand correctly that button to show logs is unreachable when error happens and makes it difficult for user to assist us with figuring out what went wrong?
2. Interface can benefit from some hint as to how to contact the team and which way is preferred

isSerge:
@nazar-pc 
1. You are right. Maybe we could have an error component as a whole page instead of modal. What do you think @achiurizo ?
2. We had a discussion on this and it will be added once all details are clarified

achiurizo:
> 1. You are right. Maybe we could have an error component as a whole page instead of modal. What do you think @achiurizo ?

Yep that sounds fine with me :+1:

src/pages/Index.vue
```diff
@@ -70,7 +70,8 @@ export default defineComponent({
     },
     async firstLoad() {
       util.infoLogger("INDEX | First Time RUN, resetting reward address")
-      await config.update({ rewardAddress: "" });
+      // reset node name in case there is a leftover from prev launch (restart due to error)
+      this.store.setNodeName(config, '');
```
ozgunozerk, line 74: nice catch

ozgunozerk (dismissed):
looking good to me, but let's wait for Nazar's approval

src/lib/client.ts
```diff
@@ -158,13 +154,14 @@ export class Client implements IClient {
...
+  public async startFarming(path: string, plotSizeGB: number): Promise<void> {
     // convert GB to Bytes
     const plotSize = Math.round(plotSizeGB * 1024 * 1024 * 1024)
     const { rewardAddress } = (await config.read());
     if (rewardAddress === "") {
       util.errorLogger("Tried to send empty reward address to backend!")
     }
-    return await tauri.invoke("farming", { path, rewardAddress, plotSize })
+
+    return tauri.invoke("farming", { path, rewardAddress, plotSize });
```
ozgunozerk, line 165: for `startFarming` and `startNode` -> you changed the signature from `Promise<boolean>` to `Promise<void>`.
We discussed it before, and if it will work like that, I'm pro for this simplification. 
However, the line `return tauri.invoke("farming", { path, rewardAddress, plotSize });` is still returning `true`/`false` (same for node), because the backend functions' signatures are the same.

isSerge: `invoke` returns `Promise<unknown>`, but we don't really care about value here, only if it resolves or not

isSerge: we don't use value in the `store.ts`: `await client.startFarming(this.plotDir, this.plotSizeGB);`

nazar-pc (approved):

Issue #297, opened 2022-07-27
In-line documentation of code


achiurizo:
Let's start backfilling in-line documentation for the codebase. Let's use [JSDoc](https://github.com/jsdoc/jsdoc) for our Vue side and document our rust backend as well.

ozgunozerk:
- [ ] vue -> @isSerge 
- [ ] rust -> @ozgunozerk

PR #303, opened 2022-08-03
Farmer simplify

isSerge:
Minor code simplification of `farm` method in `farmer.rs` as well as corresponding methods in the frontend


src-tauri/src/farmer.rs
```diff
@@ -41,16 +39,16 @@ impl Default for ArchivingFrom {
 }
 
 #[tauri::command]
-pub(crate) async fn farming(path: String, reward_address: String, plot_size: u64) -> bool {
-    // create a channel to listen for farmer errors, and restart another farmer instance in case any error
-    let (error_sender, mut error_receiver): (Sender<()>, Receiver<()>) = channel(1);
-
+pub(crate) async fn farming(
```
achiurizo, line 42: Let's start [adding documentation](https://github.com/subspace/subspace-desktop/issues/297) with our changes now. This will make it easier for us to ramp up to fuller coverage. :pray:

src/lib/client.ts
```diff
@@ -159,8 +155,7 @@ export class Client implements IClient {
   }
 
   /* FARMER INTEGRATION */
-  // TODO: method should return Promise<void>, update after backend is updated
-  public async startFarming(path: string, plotSizeGB: number): Promise<boolean> {
+  public async startFarming(path: string, plotSizeGB: number): Promise<void> {
```
achiurizo, line 158: Same here, also could start introducing [tsdoc](https://tsdoc.org/) documentation as you mentioned prior

src/stores/store.ts
```diff
@@ -272,9 +269,7 @@ export const useStore = defineStore('store', {
         this.setPlotMessage('dashboard.verifyingPlot');
         this.setNetworkMessage('dashboard.verifyingNet');
 
-        // TODO: remove variable assignment after backend (Rust) is updated
-        const isFarming = await client.startFarming(this.plotDir, this.plotSizeGB);
-        if (!isFarming) { throw Error; }
+        await client.startFarming(this.plotDir, this.plotSizeGB);
```
achiurizo, line 272: nitpick: should we normalize `this.plotDir` as `this.plotPath`? We do this with `this.plotSizeGB`

src-tauri/src/farmer.rs
isSerge: I am already adding it in another branch - `add-code-doc`

nazar-pc: In Rust it is common practice (typically enforced by lints) to require documenting all public API surface. Internal API (like this) is typically documented only if it is not trivial or requires it for some other reason. Otherwise might be just a waste of time commenting obvious things for the sake of doing it.

ozgunozerk: I agree with Nazar on this one. 
The exception maybe this: the functions that are annotated with `#[tauri::command]` are exposed to frontend, so they are not staying exactly in the crate, although they are `pub(crate)`. 
If you want, I can document all the `#[tauri::command]` ones, because they are somewhat like `public API` to some degree, but not completely...

Decision is yours :)

isSerge:
@ozgunozerk 
that is not correct, this is the method that saves `rewardAddress` and other stuff into the config file:
```
async confirmPlottingSetup(config: IConfig, util: IUtil) {
      try {
        const nodeName = util.generateNodeName();
        this.setNodeName(config, nodeName);

        await config.update({
          rewardAddress: this.rewardAddress,
          plot: {
            location: this.plotPath,
            sizeGB: this.plotSizeGB,
          },
          nodeName,
        });
      } catch (error) {
        this.setError({
          title: 'errorPage.configUpdateFailed',
          // TODO: replace default error message with specific one
          message: 'errorPage.defaultErrorMessage',
        });
      }
    },
    ```
It is called when user confirms plotting setup, before transition to `PlottingProgress` page

**Comment by ozgunozerk** [comment]
Then there was a race over this. Because before my fix, i got 60-70% of the time, `tried to sent empty reward address to farmer` error. we can undo my commit, and fix the race instead then

**Comment by ozgunozerk** [comment]
Here is the reason for the race:
```typescript
 if (!this.store.rewardAddress) {
        util.infoLogger('SETUP PLOT | reward address was empty, creating a new one');
        await util.showModal(mnemonicModal, { handleConfirm: this.handleConfirm });
      }
```

showModal is being awaited, but the result of showModal is not being awaited on. That's why `handleConfirm()` is immediately called when the `showModal` is triggered.
But in this state, user is not done with the modal. And during all this, the application proceeded into `plottingProgress` page, and tried to start node and farmer, whilst the `rewardAddress` is not yet stored in config.

ozgunozerk (dismissed):
my issues are resolved :)

src-tauri/src/farmer.rs
```diff
@@ -41,16 +39,16 @@ impl Default for ArchivingFrom {
...
+    path: String,
+    reward_address: String,
+    plot_size: u64,
+) -> Result<String, String> {
```
nazar-pc, line 46: @ozgunozerk why is it necessary to return anything other than `()` in success case?

src-tauri/src/farmer.rs
```diff
@@ -59,43 +57,39 @@ pub(crate) async fn farming(path: String, reward_address: String, plot_size: u64
             dsn_sync: false,
         };
 
-        if let Err(err) = farm(path.clone().into(), farming_args.clone()).await {
-            error!("farm function failed to start, with error: {err}");
-        }
```
nazar-pc, line 64: Well, this should have returned an error back, so you can render it on the frontend instead of moving it completely into background that will spin forever in an infinite loop printing gigabytes of errors once error happens. Do I understand that now user will have no indication that farmer isn't working at all unless they look into logs?

src-tauri/src/farmer.rs
ozgunozerk: It is not necessary for sure. 
in the frontend, at some point we may want to log the success message from the backend, that was my thought. I can remove it

src-tauri/src/farmer.rs
ozgunozerk: > Do I understand that now user will have no indication that farmer isn't working at all unless they look into logs?

it is actually the same case as before, nothing changed about that. 
1. Frontend is invoking the `startFarming` function, and backend gives back a result regarding the farmer started successfully or not. 
2. then, somehow if farming crashes (due to node restart for example), we are trying to restart the farmer. If these will fail continuously, yes, frontend won't have a direct way of knowing it. 

Till now, frontend only knew about the first start of the farmer succeeded or not (and it is still like that).

nazar-pc: Yes, but you could have replaced `if let Err` with `?` and it would return an error, now everything happens in the background instead, so no way to catch start failure.

ozgunozerk: I see your point. 
I will work on it. In the meantime, we will still not be able to tell frontend if the subsequent farmer instances fail. If we want to solve this one too, maybe another approach will be required.

ozgunozerk: how about this?
https://github.com/subspace/subspace-desktop/blob/farmer-simplify-suggestions/src-tauri/src/farmer.rs#L60-L88

nazar-pc: Yes, just add some message to `error!("{error}")` so it is easier to find context.

src-tauri/src/farmer.rs
ozgunozerk: @isSerge , your take on this? will it be helpful for frontend to provide a success message, or no?

src-tauri/src/farmer.rs
```diff
@@ -59,43 +57,49 @@ pub(crate) async fn farming(path: String, reward_address: String, plot_size: u64
...
-            error!("farm function failed to start, with error: {err}");
+        let farming_handle;
+        match farm(path.clone().into(), farming_args.clone()).await {
+            Err(error) => {
+                return Err(format!(
+                    "farm function failed to start, with error: {error}"
+                ))
+            }
+            Ok(handle) => farming_handle = handle,
         }
```
nazar-pc, line 68:
```suggestion
        let farming_handle = farm(path.clone().into(), farming_args.clone())
            .await
            .map_err(|error| format!(
                "farm function failed to start, with error: {error}"
            ))?;
```

nazar-pc (approved):

src-tauri/src/farmer.rs
ozgunozerk: thanks! will commit after Serge replies, maybe I will change it to `Result<(), String>` as you suggested,

src-tauri/src/farmer.rs
isSerge: success message - not really, error maybe, but not critical at the moment

nazar-pc (approved):

ozgunozerk:
@nazar-pc seems like expected checks needs to be configured for this one as well

nazar-pc:
No, this one needs to be updated according to new workflow.

PR #304, opened 2022-08-03
Workflow update (abandon tauri actions)

### LINKED ISSUE #255: Potential User Confusion on Which Installer to Download
Adding which filetype corresponds to which operating system when downloading subspace-desktop would be helpful.

### LINKED ISSUE #275: Create a new publish workflow
Soon, we may have 2 separate executables, the second one will be for `opencl` compatibility for the ones who want to utilize their GPUs. 

Right now, it is not possible to have more than 1 executable per OS using `tauri-action` yaml workflow. Because we would need to call the action 2 times, and each call checks for a unique `tag`/`release name`. If the `tag` exists, the workflow is programmed to fail. 

So, we should write a new worklflow, and abandon `tauri-actions`.

With the new workflow, written by us, and more low-level, we will have more control over the build process on the CI, and will be able to solve #255 as well :)

ozgunozerk:
Fixes #255 
Fixes #275 

We finally have a new workflow, which allows us to have more granular control over the build process and everything related to executables. The biggest reason was, to be able to publish 2 versions (1 for normal, 1 with opencl).

Note:
Cargo toml changes might seem weird for a workflow PR -> the reason is cirrus-runtime inclusion bug had to be solved on the monorepo side

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -3,9 +3,11 @@ on:
   push:
     paths-ignore:
       - "**.md"
+      - "**/publish.yml"
   pull_request:
     paths-ignore:
       - "**.md"
+      - "**/publish.yml"
 
 env:
   # Not needed in CI, should make things a bit faster
@@ -21,7 +23,7 @@ jobs:
       matrix:
         os:
           - ubuntu-22.04
-          - macos-11
+          - macos-12
           - windows-2022
 
     runs-on: ${{ matrix.os }}
@@ -109,7 +111,7 @@ jobs:
       matrix:
         os:
           - ubuntu-22.04
-          - macos-11
+          - macos-12
           - windows-2022
 
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -132,37 +132,13 @@ jobs:
         env:
           TAURI_PRIVATE_KEY: ${{ secrets.TAURI_PRIVATE_KEY }}
           TAURI_KEY_PASSWORD: ${{ secrets.TAURI_KEY_PASSWORD }}
-
-      - name: Sign Application (macOS)
-        if: runner.os == 'macOS'
-        run: |
-          echo "Importing certificate"
-          echo "${{ secrets.MACOS_CERTIFICATE }}" | base64 --decode > certificate.p12
-          security create-keychain -p "${{ secrets.MACOS_CERTIFICATE_PW }}" build.keychain
-          security default-keychain -s build.keychain
-          security unlock-keychain -p "${{ secrets.MACOS_CERTIFICATE_PW }}" build.keychain
-          security import certificate.p12 -k build.keychain -P "${{ secrets.MACOS_CERTIFICATE_PW }}" -T /usr/bin/codesign
-          security set-key-partition-list -S apple-tool:,apple:,codesign: -s -k "${{ secrets.MACOS_CERTIFICATE_PW }}" build.keychain
-          echo "Signing farmer"
-          codesign --force --options=runtime --entitlements .github/workflows/Entitlements.plist -s "${{ secrets.MACOS_IDENTITY }}" --timestamp ${{ env.PRODUCTION_TARGET }}/dmg/subspace-desktop_0.6.10_aarch64.dmg
-          echo "Creating an archive"
-          mkdir ${{ env.PRODUCTION_TARGET }}/macos-installer
-          cp ${{ env.PRODUCTION_TARGET }}/dmg/subspace-desktop_${{ github.ref_name }}_aarch64.dmg ${{ env.PRODUCTION_TARGET }}/macos-installer
-          ditto -c -k --rsrc ${{ env.PRODUCTION_TARGET }}/macos-installer subspace-installer.zip
-          echo "Notarizing"
-          xcrun altool --notarize-app --primary-bundle-id binaries-${{ github.ref_name }} --username "${{ secrets.MACOS_APPLE_ID}}" --password "${{ secrets.APPLE_PASSWORD }}" --file subspace-installer.zip
-          # TODO: Wait for notarization before stapling
-          # echo "Stapling installer"
-          # xcrun stapler staple ${{ env.PRODUCTION_TARGET }}/dmg/subspace-desktop_${{ github.ref_name }}_aarch64.dmg
-          echo "Done!"
-
-      - name: Sign Application (Windows)
-        uses: skymatic/code-sign-action@v1.1.0
-        with:
-          certificate: ${{ secrets.WINDOWS_CERTIFICATE }}
-          password: ${{ secrets.WINDOWS_CERTIFICATE_PW }}
-          certificatesha1: 00A427587B911908F59B6C42BA2863109C599C1C
-          folder: ${{ env.PRODUCTION_TARGET }}/msi
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          ENABLE_CODE_SIGNING: ${{ secrets.MACOS_CERTIFICATE }}
+          APPLE_CERTIFICATE: ${{ secrets.MACOS_CERTIFICATE }}
+          APPLE_CERTIFICATE_PASSWORD: ${{ secrets.MACOS_CERTIFICATE_PWD }}
+          APPLE_SIGNING_IDENTITY: ${{ secrets.MACOS_IDENTITY_ID }}
+          APPLE_ID: ${{ secrets.APPLE_ID }}
+          APPLE_PASSWORD: ${{ secrets.APPLE_PASSWORD }}
 
       - name: Prepare installers for uploading (Ubuntu)
         run: |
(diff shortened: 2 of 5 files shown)


.github/workflows/publish.yml
```diff
@@ -57,28 +102,88 @@ jobs:
...
+        with:
+          name: executables-${{ matrix.build.suffix }}
+          path: |
+            executables/*
+          if-no-files-found: error
+
+      - name: Upload executables to assets
+        uses: alexellis/upload-assets@0.3.0
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
```
achiurizo, line 186: should this also be `secrets.GITHUB_TOKEN`? If so, perhaps we should standardize the usage (either this way or the other)

https://github.com/subspace/subspace-desktop/pull/304/files#diff-551d1fcf87f78cc3bc18a7b332a4dc5d8773a512062df881c5aba28a6f5c48d7L70

.github/workflows/ci-tests.yml
```diff
@@ -21,7 +23,7 @@ jobs:
       matrix:
         os:
           - ubuntu-22.04
```
achiurizo, line 25:
```suggestion
          - ubuntu-20.04
```

do we want to match this with what we have in [publish.yml](https://github.com/subspace/subspace-desktop/pull/304/files#diff-551d1fcf87f78cc3bc18a7b332a4dc5d8773a512062df881c5aba28a6f5c48d7R16)

ozgunozerk: I believe we don't. It was like that in monorepo as well. I guess for testing, we are aiming for the latest technology, but for production, the most compatible one. That was my take

.github/workflows/publish.yml
ozgunozerk: yeah it is probably `secrets.GITHUB_TOKEN`, let me push that change

achiurizo (approved):

ImmaZoni (dismissed):
LGTM

.github/workflows/publish.yml
```diff
@@ -1,21 +1,55 @@
...
+      RUSTFLAGS: ${{ matrix.build.rustflags }}
 
-    runs-on: ${{ matrix.os }}
     steps:
+      - name: Maximize build space
+        if: runner.os == 'Linux'
+        run: |
+          sudo rm -rf /usr/share/dotnet
+          sudo rm -rf /opt/ghc
+          sudo rm -rf "$AGENT_TOOLSDIRECTORY"
```
nazar-pc, line 45: Are these still necessary? I'm worried this is just a temporary solution.

Fine for now, but we need to find longer-term solution somehow.

.github/workflows/publish.yml
```diff
@@ -31,18 +65,29 @@ jobs:
...
+        if: runner.os == 'Linux'
+
+      # Workaround to resolve link error with C:\msys64\mingw64\bin\libclang.dll
       - name: Remove msys64
         run: Remove-Item -LiteralPath "C:\msys64\" -Force -Recurse
         if: runner.os == 'Windows'
 
-      - name: install webkit2gtk (ubuntu only)
-        if: runner.os == 'Linux'
+      # libayatana does not work for ubuntu20.04 -> https://github.com/subspace/subspace-desktop/runs/7581578859?check_suite_focus=true
```
nazar-pc, line 85: Why not? I have sent you a commit a few days ago that fixes it by simply updating a dependency. Specifically this one: https://github.com/subspace/subspace-desktop/commit/b4a96bea91157422cb5987f1f5a751ad9872d6cf

Legacy `libappindicator3` conflicts with `libayatana-appindicator3-1` and will likely make users choose either they have Subspace Desktop or other apps.

.github/workflows/publish.yml
```diff
@@ -57,28 +102,88 @@ jobs:
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
-          tagName: __VERSION__ # the action automatically replaces \_\_VERSION\_\_ with the app version
-          releaseName: "__VERSION__"
-          releaseBody: "See the assets to download this version and install."
-          releaseDraft: false
-          prerelease: true
+          asset_paths: '["executables/*"]'
+        if: github.event_name == 'push' && github.ref_type == 'tag'
```
nazar-pc, line 189: I believe either one or the other or both will fail unless you have proper `permissions` section in the workflow. Please check monorepo, it is there for a reason.

.github/workflows/publish.yml
```diff
@@ -1,21 +1,55 @@
...
+        run: |
+          sudo rm -rf /usr/share/dotnet
+          sudo rm -rf /opt/ghc
+          sudo rm -rf "$AGENT_TOOLSDIRECTORY"
       - name: git checkout
         uses: actions/checkout@v2
 
+      - name: Extract version
+        id: extract_version
+        uses: Saionaro/extract-package-version@v1.0.6
```
nazar-pc, line 51: Wouldn't it be better to use tag as a version instead? It will be the same in most cases, but you could also do nightly builds or other interesting things and still have executabled being called uniquely instead of all look the same. Just a thought.

nazar-pc (changes requested):
> Cargo toml changes might seem weird for a workflow PR -> the reason is cirrus-runtime inclusion bug had to be solved on the monorepo side

It didn't necessarily have to be in the same PR though.

.github/workflows/publish.yml
ozgunozerk: Hmm, I thought since we are not planning to use dotnet and ghc, it could be a permanent solution...
I think it is still necessary, I can double-check in a fork, and look for other solutions if you like

.github/workflows/publish.yml
ozgunozerk: In the monorepo, the `permissions` field is only there for the `container-linux` job (https://github.com/subspace/subspace/blob/main/.github/workflows/snapshot-build.yml#L22), not present for building the executables. Am I missing something here?

.github/workflows/publish.yml
ozgunozerk: yes this was a remnant of the previous strategy for uploading the executables to assets, will change

ozgunozerk: Oh actually no, just remembered:
when tauri builds the executable, it automatically appends the version string to the executable name. So we have to get the version from the package.json for this workflow. I can't remove that.

However, if your suggestion was not to get rid of this, but only to use tag as the version, I can do that

.github/workflows/publish.yml
nazar-pc: Well, yes, but it only frees some space for us. If we fit into unmodified I'd prefer to not delete anything, but not blocking of course.

.github/workflows/publish.yml
nazar-pc: Hm, I guess we can use tag for the final name of the file and leave whatever Tauri generates as source. BTW, for file copying you can use `*` and not exact match, thus potentially avoiding bothering with specific versions.

.github/workflows/publish.yml
nazar-pc: Hm, either it affects other jobs anyway or it is indeed only necessary for container images and I got it mixed in my head. If it worked in test runs then ignore this comment.

ozgunozerk: In the test runs, I did not create any release or pushed a tag, that's why this step is actually not tested. 
I don't think the the permissions set for one job will affect the other job, but that's only my guess :)
Is it ok to test this via creating a pre-release and then deleting that pre-release? If not, I will take a note of this, and in case of this permission error, we can quickly update the workflow

nazar-pc: I was testing such things in my personal fork. You can create as many test releases from whatever branch as you'd like. But let's no do such experiments in official repos if possible.

.github/workflows/publish.yml
ozgunozerk: There are other files in the directory that I'm copying the bundle, `*` did not seem safe to me... 
This extra step is very minimal and does not take much time. I believe it makes the workflow safer (we have access to the full filename)

.github/workflows/publish.yml
nazar-pc: Actually yes, I am quite certain it'll fail (you'll see familiar nicknames there): https://github.com/alexellis/upload-assets/issues/6

Probably GitHub uses the same token for multiple jobs in the same workflow or something, so it accidentally worked in monorepo.

.github/workflows/publish.yml
nazar-pc: Sure. But you don't have to use just `*`, you can use `target/*/*.msi` for instance and capture file `target/x86-64.../something.msi`. Anyway, works fine as is too.

.github/workflows/publish.yml
ozgunozerk: `permissions` were not required -> https://github.com/ozgunozerk/subspace-desktop/releases/tag/hoping
the errors they are getting the issue you linked, I have faced it as well. Maybe they did the same mistake as me (assumed that action would create the release).

nazar-pc (approved):

PR #305, opened 2022-08-05
fix incorrect error popup for farmer restart

ozgunozerk:
Here is the fix for the scenario: when farmer is restarted, an error page appears, saying that farmer could not be restarted. 

The issue is simple, but I will explain them in comments, please see below.

Draft, because I have better ideas on how to handle this, and probably you can do a much better job than me Serge. This is just to point out how to fix the error

### DIFF
--- a/src/stores/store.ts
+++ b/src/stores/store.ts
@@ -269,15 +269,25 @@ export const useStore = defineStore('store', {
         this.setNetworkMessage('dashboard.verifyingNet');
 
         await client.startFarming(this.plotPath, this.plotSizeGB);
-
         // TODO: consider moving logging to client.ts
-        util.infoLogger('PLOTTING PROGRESS | farmer started');
+        util.infoLogger('Farmer started');
+      } catch (error) {
+        // TODO: consider moving logging to client.ts
+        util.errorLogger('Farmer start error!');
+        this.setError({
+          title: 'errorPage.startFarmerFailed',
+          // TODO: replace default error message with specific one
+          message: 'errorPage.defaultErrorMessage',
+        });
+        return
+      }
 
-        const syncState = await client.getSyncState();
-        this.setSyncState(syncState);
-        let isSyncing = await client.isSyncing();
+      const syncState = await client.getSyncState();
+      this.setSyncState(syncState);
+      let isSyncing = await client.isSyncing();
 
-        do {
+      do {
+        try {
           await new Promise((resolve) => setTimeout(resolve, 3000));
           const syncState = await client.getSyncState();
           this.setSyncState(syncState);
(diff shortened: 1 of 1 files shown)


src/stores/store.ts
```diff
@@ -269,15 +269,25 @@ export const useStore = defineStore('store', {
         this.setNetworkMessage('dashboard.verifyingNet');
 
         await client.startFarming(this.plotPath, this.plotSizeGB);
-
         // TODO: consider moving logging to client.ts
-        util.infoLogger('PLOTTING PROGRESS | farmer started');
+        util.infoLogger('Farmer started');
+      } catch (error) {
```
ozgunozerk, line 274: first of all, this catch is printing the error of "farmer start error". 
So, the scope of the `try` block should only encapsulate the `farmer start` functionality. 

In the previous version, where `try` block is containing the whole logic, the error was happening after the farmer start is successful, yet the printed error is `farmer start error`, this is confusing. 

Hence, I narrowed down the scope of the `try` block here.

src/stores/store.ts
```diff
@@ -269,15 +269,25 @@ export const useStore = defineStore('store', {
...
-        let isSyncing = await client.isSyncing();
+      const syncState = await client.getSyncState();
+      this.setSyncState(syncState);
+      let isSyncing = await client.isSyncing();
 
-        do {
+      do {
+        try {
           await new Promise((resolve) => setTimeout(resolve, 3000));
           const syncState = await client.getSyncState();
```
ozgunozerk, line 292: this is where the error happens -> `getSyncState()`. 

The reason is, farmer/node is going to be restarted, and at the moment not available. Hence, the `getSyncState()` is failing because the WebSocket is closed. 

My solution was to wrap these with `try`, and simply print&forget the error. 

What can be better is: try for like 20-30 seconds if it is creating an error, and if within 20-30 seconds, it can continue without errors, that's fine. If the error is still continuing after 20-30 seconds, then show a related error page, and ask for a restart.

My current solution does not handle this 20-30 seconds of `trial window`. It only logs the error and keeps trying again. Not the best solution, but way better than what we had before

isSerge:
Maybe we simply should not render Dashboard or PlottingProgress while node and farmer are restarting. Instead, we can display the loading page the same as on restart. 

I think store logic is correct - we should not farm if we can't get `syncState`, we can definitely add specific try-catch blocks if we want to identify a particular step when an error occurred, but probably not right now.

ozgunozerk:
> Maybe we simply should not render Dashboard or PlottingProgress while node and farmer are restarting. Instead, we can display the loading page the same as on restart.

the backend cannot call the frontend functions. So when a new node name is set, the application should stop the syncing part of the farming function (the one that is continuing on the frontend, not the backend) and render a loading page for 5-10 seconds. 

But i guess all these would complicate the application both for us and for the users.

> I think store logic is correct - we should not farm if we can't get syncState, we can definitely add specific try-catch blocks if we want to identify a particular step when an error occurred, but probably not right now.

If the farmer started successfully, that means in the frontend -> node was successfully started, and farmer is fine. Then if we encounter a `syncState` error, that means node is restarting, and we can wait for 20-30 seconds, and during this time, ignore the errors. That was my suggestion, and I think it is the most simple way of not distracting the users throughout this restart process, and hide it as a backend complexity

isSerge:
I think it's simpler if UI reflects the state of node and farmer without extra assumptions, also I am not a big fan of ignoring errors - might be harder to debug if smth is wrong. I believe we can utilize [tauri event module](https://tauri.app/v1/api/js/modules/event/#event) to let UI know when node has restarted

ozgunozerk:
Yes that should be the ultimate goal, i agree. 
Let's focus on this one after finishing our current tasks then.

PR #307, opened 2022-08-08
Add code documentation

isSerge:
The initial purpose was to just add documentation comments, but some modules required refactoring in order to be documented (more refactoring is needed, but I think it's better than before). Some unit tests were added to make sure refactoring did not break stuff.


achiurizo (approved):

ozgunozerk (approved):
Lots of changes :)
Better structure than before, thanks!
Don't know for sure it is bug free though (hoping it won't introduce any bug regarding autolaunch and config). 
Good that we are having incrementally more tests, hopefully we won't have to worry about new bug introductions if they are failing the tests

PR #309, opened 2022-08-09
add unit tests run into workflow

isSerge:


ozgunozerk:
I don't think we need unit tests in publish.yml

In general, i did not put any tests in publish, it is just for publishing

ozgunozerk (approved):

PR #310, opened 2022-08-10
Add getSyncState retry

### LINKED ISSUE #306: False-positive error page on frontend when node is restarted
After new error pages and node restart functionality, we have a new bug:

- When the node is restarted, the restart process is taking a bit of time.
- During this restart process, the frontend still tries to query the websocket via `getSyncState()`
- because the node is not available during that time, the query fails, and an error page is displayed with a restart option
- however, everything in the backend is working as expected
- if frontend can know about this, and waits for a few seconds, or waits for a signal from the backend, then frontend can stop querying the websocket. 

Here is the further discussion about this: https://github.com/subspace/subspace-desktop/pull/305

isSerge:
This is a temporary workaround for #306 while I will be working on a better long-term solution, which will require replacing loop with subscription. The latter can be stopped before the node is restarted and resumed after API is ready

---
closes #306

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "@vueuse/core": "^6.0.0",
     "apexcharts": "^3.27.3",
     "date-fns": "^2.28.0",
+    "p-retry": "^5.1.1",
     "pinia": "^2.0.16",
     "process": "^0.11.10",
     "quasar": "^2.7.5",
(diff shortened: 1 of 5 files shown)


ozgunozerk (approved):
considering default trial is 10 attempts, and there is 1 sec interval between each of the attempts, sounds fine to me!

achiurizo (approved):

Issue #308, opened 2022-08-08
Error page: add specific error resolutions


isSerge:
<img width="800" alt="image" src="https://user-images.githubusercontent.com/13568875/182025295-7699b0bf-d1b8-4644-9062-f56ef2864abf.png">

The current error page body has a generic message, we need to add specific resolutions for these cases: 
- could not update or read the config file
- could not start node
- could not start farming

Also, it will be nice to add more details on how to contact the team

ozgunozerk:
For the `startFarmer` function in `store.ts` -> I think the error handling there is misleading and I suggest some changes.

*Background info:*

- Lines: https://github.com/subspace/subspace-desktop/blob/main/src/stores/store.ts#L316-L324
are responsible from the successful `farmer` start on the tauri/backend side. If an error happens in here, it means there is a problem in the backend related to `farming` function.

- Lines: https://github.com/subspace/subspace-desktop/blob/main/src/stores/store.ts#L316-L324
are responsible from the syncing process, if a problem happens in these lines, it means that a problem happened between the farmer-node connection, and this connection was related to syncing process

- Lines: https://github.com/subspace/subspace-desktop/blob/main/src/stores/store.ts#L316-L324
are responsible from starting a successful block subscription, and if a problem happens, it means there is problem between the node and farmer again, but this time, for block subscription.

## Why it is misleading:
all these 3 possible errors are handled with a single error message, namely "farmer start error". And this is misleading, because in the 2nd and the 3rd scenario I described above, the farmer INDEED started successfully. This error message will confuse both the users and us (the developers). 

2 solutions comes to my mind:
1- have a separate `try/catch` block for all of them, and in the `catch` block, handle the respective error with a relevant error message
2- separate the 2nd and the 3rd scenarios into another function, and handle the errors in those functions.

PR #318, opened 2022-08-22
Upgrade dependencies

ozgunozerk:
There was a bug with `upgrade-dependencies` #314 
This PR updates all the dependencies once more to the latest main of the monorepo. Issue seems to be resolved with it :)


src-tauri/src/farmer.rs
```diff
@@ -203,12 +234,49 @@ async fn farm(
     )
     .await?;
 
-    Ok(multi_plots_farm)
-}
+    // Start RPC server
+    let ws_server = match WsServerBuilder::default()
+        .build(ws_server_listen_addr)
+        .await
+    {
```
nazar-pc, line 241: Why do you need farmer's RPC server?

nazar-pc (commented):
I have changed the target branch to see new changes more clearly, feel free to change back if needed

src-tauri/src/farmer.rs
ozgunozerk: we don't. I was not putting it in here since farmer DSN is not ready till now. This time I put it in because every update I have to delete all the stuff related to farmer dsn in here, and thought if it won't hurt, let's include it, it will make the updates easier.

On the other hand, if you suggest to remove it, im fine with it, I can remove them

src/stores/store.ts
```diff
@@ -321,16 +321,26 @@ export const useStore = defineStore('store', {
         await client.startFarming(this.plotPath, this.plotSizeGB);
 
         // TODO: consider moving logging to client.ts
-        util.infoLogger('PLOTTING PROGRESS | farmer started');
+        util.infoLogger('farmer started');
 
+      } catch (error) {
```
isSerge, line 326: I don't really get why adding one more `try-catch` if error handling is the same for both

src/stores/store.ts
```diff
@@ -359,14 +369,14 @@ export const useStore = defineStore('store', {
...
         // TODO: consider moving logging to client.ts
-        util.infoLogger('PLOTTING PROGRESS | block subscription started');
+        util.infoLogger('block subscription started');
       } catch (error) {
         // TODO: consider moving logging to client.ts
-        util.errorLogger('PLOTTING PROGRESS | Farmer start error!');
+        util.errorLogger('Sync error: ' + error as string);
         this.setError({
-          title: 'errorPage.startFarmerFailed',
+          title: 'errorPage.startFarmerFailed', // TODO: make a page for sync error
```
isSerge, line 377: We don't need another page, currently, we have one component for all errors with different titles and messages

isSerge: I would suggest we don't make these changes in `store.ts` unless we clarify error resolutions #308

nazar-pc (approved):

PR #321, opened 2022-08-23
rotating logs

ozgunozerk:
@i1i1 , I had to tinker with `BunyanFomatter` code, I checked and it is working as I expected, but double checking on everything related to logging is alright on Linux side would be helpful :)

Since tokio-tracing team are not responsive to the PR anymore (don't know why), I will probably maintain this fork

### DIFF
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -12,7 +12,7 @@ mod node;
 mod utils;
 
 use anyhow::Result;
-use std::fs::{create_dir_all, File};
+use std::fs::create_dir_all;
 use tauri::SystemTrayEvent;
 use tauri::{Manager, RunEvent, WindowEvent};
 use tracing::level_filters::LevelFilter;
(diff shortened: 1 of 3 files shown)


src-tauri/src/main.rs
```diff
@@ -21,13 +21,18 @@ use tracing_subscriber::prelude::*;
 use tracing_subscriber::Layer;
 use tracing_subscriber::{fmt, fmt::format::FmtSpan, EnvFilter};
 
+const KEEP_LAST_N_DAYS: usize = 5;
```
i1i1, line 24: Do we need 5 days though? I think keeping just one day would be more than enough. You can increase level of logs though and keep smaller time span

nazar-pc: I'd rather say 14 days would be preferred, depending on how big the log is

ozgunozerk: i recall the daily logs wont reach GBs. It might be around 10MB maybe, but i'm not 100% sure.

ozgunozerk: I've done some testing, approx 1MB per hour, so 24MB per day. 
What is our final decision on this parameter in this case?

PR #325, opened 2022-08-23
new chainspec

ozgunozerk:


nazar-pc (commented):
1) It is already outdated
2) Do we really want to merge it into `main`? It is not supposed to be in the main repo.

ozgunozerk:
Oh... I believe Ivan sent the link for the chainspec, I did not double check and just copied it. 
In the product meeting, it has been decided to include the new chainspec as soon as we can, and hopefully make it ready by tomorrow night. 
If we do not want to put it in main, then I guess we don't need a PR, we can build the executables from any fork.

I'm currently away from the computer. (@achiurizo, just notifying you about the updates)

nazar-pc:
It was correct, but I found an issue with the version we've tested, so there is an update in the branch we were building it from.

PR #327, opened 2022-08-30
rotating logs and cargo toml cleanup

ozgunozerk:
Previous PR had conflicts with main regarding cargo lock, it was easier to create a new branch on top of the up to date main. 
I made the parameter 7 days for keeping the logs (it was about 24-30 mb daily for my case)

### DIFF
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -70,7 +70,9 @@ codegen-units = 1
 # TODO: Remove once chacha20poly1305 0.10 appears in libp2p's dependencies
 chacha20poly1305 = { git = "https://github.com/RustCrypto/AEADs", rev = "06dbfb5571687fd1bbe9d3c9b2193a1ba17f8e99" }
 libp2p = { git = "https://github.com/subspace/rust-libp2p", branch = "subspace-v3" }
-tracing = { git = "https://github.com/ozgunozerk/tracing-keep-last-n-logs", branch = "v0.1.x" }
-tracing-core = { git = "https://github.com/ozgunozerk/tracing-keep-last-n-logs", branch = "v0.1.x" }
-tracing-appender = { git = "https://github.com/ozgunozerk/tracing-keep-last-n-logs", branch = "v0.1.x" }
-tracing-subscriber = { git = "https://github.com/ozgunozerk/tracing-keep-last-n-logs", branch = "v0.1.x" }
+# if https://github.com/tokio-rs/tracing/pull/2284 merges, we can get rid of the below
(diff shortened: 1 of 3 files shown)


i1i1:
@ozgunozerk in those cases you can try deleting `Cargo.lock` and just run `cargo check`. This should generate a new one.

Though it has downsides as it will update all dependencies

i1i1 (dismissed):
LGTM. If the size of logs is our concern, may be a good idea would be to implement some lz4 storage layer which would compress logs.

This might allow us to increase log level for some crates (I think at least for `subspace` target at least `debug` should be helpful).

src-tauri/Cargo.toml
```diff
@@ -71,3 +70,7 @@ codegen-units = 1
 # TODO: Remove once chacha20poly1305 0.10 appears in libp2p's dependencies
 chacha20poly1305 = { git = "https://github.com/RustCrypto/AEADs", rev = "06dbfb5571687fd1bbe9d3c9b2193a1ba17f8e99" }
 libp2p = { git = "https://github.com/subspace/rust-libp2p", branch = "subspace-v3" }
+tracing = { git = "https://github.com/ozgunozerk/tracing-keep-last-n-logs", branch = "v0.1.x" }
+tracing-core = { git = "https://github.com/ozgunozerk/tracing-keep-last-n-logs", branch = "v0.1.x" }
+tracing-appender = { git = "https://github.com/ozgunozerk/tracing-keep-last-n-logs", branch = "v0.1.x" }
+tracing-subscriber = { git = "https://github.com/ozgunozerk/tracing-keep-last-n-logs", branch = "v0.1.x" }
```
nazar-pc, line 76: I'd leave a comment on why this is necessary and move forks into our repo, so we can keep it forever and make sure we can build corresponding commits in the future.

nazar-pc (commented):
> @ozgunozerk in those cases you can try deleting Cargo.lock and just run cargo check. This should generate a new one.

That will result in a significantly different lock file. Rather accepting old lock file when resolving conflicts and then amending commit after cargo check.

nazar-pc (approved):

ozgunozerk:
> LGTM. If the size of logs is our concern, may be a good idea would be to implement some lz4 storage layer which would compress logs.
> 
> This might allow us to increase log level for some crates (I think at least for `subspace` target at least `debug` should be helpful).

I agree, let us discuss this at the desktop meeting, and if we can arrive at a decision, we can take a look at it when we go back to desktop. I think it will be an overkill for now, since we may take a break from it soon

i1i1 (approved):

Issue #336, opened 2022-09-22
Trail of Bits audit - Issue #2 (insufficient validation on importing reward address)


ozgunozerk:
Right now, frontend accepts any valid hex format for the reward address, whilst the backend only accepts `sr25519` formatted reward addresses. 

One approach to solve this may be: exposing the backend validating function to frontend, so frontend can use it.

PR #341, opened 2022-09-22
reward address validation

ozgunozerk:
Solves #336 

Few words explaining this PR, it will be wordy but worth reading.

I couldn't find a good function to import from `polkatod` which will validate the reward address in `sr25519` format. 

So, I exported the backend function to validate that (we were already doing the validation on the backend). However, the exposed functions are always `async` because of `tauri invoke`. 

This new validation function being `async`, had caused more troubles than I expected, and had to create new extra variables to get around the problems. 
@isSerge please review the new code, maybe the same functionality can be achieved with simpler logic, but however I tried, it was causing some minor bugs in places. That is the best I can come up with.


src/pages/ImportKey.vue
```diff
@@ -9,58 +9,71 @@ q-page(padding)
...
   },
+  data() {
+    return {
+      rewardAddress: "",
+      validAddress: false,
+    };
+  },
   methods: {
-    isValidSubstrateAddress(val: string): boolean {
+    async isValidSubstrateAddress(addr: string): Promise<boolean|string> {
```
isSerge, line 58: I think it would be easier to keep the original function and just replace `encodeAddress` + `decodeAddress` with `tauri.invoke` without adding local state and watcher

src/pages/ImportKey.vue
```diff
@@ -9,58 +9,71 @@ q-page(padding)
...
       try {
-        encodeAddress(isHex(val) ? hexToU8a(val) : decodeAddress(val));
-        return true;
+        let result: boolean = await tauri.invoke("validate_reward_address", { addr });
```
isSerge, line 60: this can be `const` instead of `let` since there is no re-assignment

src/pages/ImportKey.vue
```diff
@@ -9,58 +9,71 @@ q-page(padding)
...
+        this.validAddress = result;
+        return (result || this.$t('importKey.rewardAddress'));
       } catch (error) {
-        return false;
+        errorLogger(error);
+        return (false || this.$t('importKey.rewardAddress'));
       }
     },
     async importKey() {
+      this.store.rewardAddress = this.rewardAddress
```
isSerge, line 69: it's better to use a method from `store` (smth like `setRewardAddress`) instead of assignment

src/i18n/en.json
```diff
@@ -86,10 +86,9 @@
   "importKey": {
     "tooltip": "Please provide a valid Reward Address to proceed.",
     "pageTitle": "Import Reward Address",
-    "rewardAddress": "Please enter your Reward Address in SS58 format (it should start with the letters \"st\")",
+    "rewardAddress": "Please enter your Reward Address in SS58 format!",
```
isSerge, line 89: I think most ppl don't know what is SS58, so I would keep `(it should start with the letters \"st\")`, alternatively we could add a link somewhere

src/pages/ImportKey.vue
ozgunozerk: The problem is: `tauri.invoke()` is async, this function needs to async as well. Then this async function is creating bugs in other places (tooltip, and error message display for the textbox). That's why we need an extra variable to watch on to get around `async`ness.

src/i18n/en.json
ozgunozerk: It does not have to start with `st` though, any valid `SS58` would do, that's why I wanted to get rid of `st`. 
Maybe we can just say it's in the wrong format, but then the user wouldn't know which format to follow and how to fix the invalid form. 

Maybe saying `it should be in "st" format` will be okay although it is not necessary. It should be a clear tip to the user.
I want to get @nazar-pc 's opinion on that though

nazar-pc: We already told the community that any address should work and I feel like we should stick to that until we hit mainnet at least.

I also don't think we need to mention SS58 in any way to the end user, we should just say something like "proper wallet address from Subwallet or Polkadot.js extension" (users that use Talisman or something will know it is still the same address and is compatible),

ozgunozerk: > "proper wallet address from Subwallet or Polkadot.js extension"
 
we can go with it if that's ok with you @isSerge

isSerge: Totally

src/pages/ImportKey.vue
isSerge: async validation should be trivial. I will take a closer look locally and add my commits

ozgunozerk:
Wonderful, thanks @isSerge

Issue #340, opened 2022-09-22
Trail of Bits - Code Quality Recommendations


ozgunozerk:
- [x]  Unnecessary complexity for (9999 - 1000 + 1) to 9000 (ozgun)
- [x]  Confusing function name (`checkDev()`), which actually disables contextmenu) (ozgun)
- [x]  Bad default directory (changing plot folder to a specific folder will fix this) (natacha)
- [x]  Duplicate code for config creation (ozgun)
- [x]  Outdated documentation ([architecture.md](http://architecture.md/)) (ozgun)
- [x]  Useless assignment (subspace-desktop/src/stores/store.ts#L330) (serge)

ozgunozerk:
PR automatically closed this one, but there is a remaining task in here. So, reopening :)

Issue #339, opened 2022-09-22
Trail of Bits audit - Issue #5 (Insufficient privilege separation between the frontend and backend)


ozgunozerk:
TL;DR: Frontend has too much power.

The following privileges granted to frontend in the `tauri.conf.json` need to be narrowed down, or completely disabled:

1. ```
    "fs": {
        "all": true,
        "scope": ["$HOME/*"]
      }

2. ``` 
    "shell": {
        "execute": true,
        "open": false,
        "scope": [
          {
            "name": "run-osascript",
            "cmd": "osascript",
            "args": true
          }
        ]
      }

- [ ] 1 (filesystem)
- [x] 2 (apple shell script)

PR #342, opened 2022-09-23
Issue4 and architecture fixes

### LINKED ISSUE #338: Trail of Bits audit - Issue #4 (Flawed regex in tauri.conf.json)
our current regex  `"open": "[/subspace\\-desktop/]",` will accept any character existing in `subspace-desktop` string. This is not we want. 

Ultimately, we shouldn't enable this, and let backend do this work to not give frontend too much privilege

ozgunozerk:
We were using `shell.open()` on the frontend to open the log folder (via default OS file explorer). 
This was issue#4 from the audit. 

I made a backend function to handle this, and disabled `shell -> open` access for the frontend. This should fix #338 , and also help #339.

Along the way: 
- I fixed the typos and outdated stuff in the `architecture.md` file, this was mentioned in #340 , so we can complete that sub-task in that issue after this PR merges.
- fixed (9999 - 1000 + 1) 
- fixed weirdly named `checkDev()`
- Duplicate code for config creation

### DIFF
--- a/src-tauri/src/utils.rs
+++ b/src-tauri/src/utils.rs
@@ -1,5 +1,6 @@
 use serde::Serialize;
 use std::path::PathBuf;
+use std::process::Command;
 use tauri::{api, Env};
 use tracing::{debug, error, info};
 
@@ -7,6 +8,27 @@ use tracing::{debug, error, info};
 pub(crate) struct DiskStats {
     free_bytes: u64,
     total_bytes: u64,
+}
+
+#[tauri::command]
+pub(crate) fn open_folder(dir: String) {
+    #[cfg(target_os = "windows")]
+    Command::new("explorer")
+        .arg(dir)
+        .spawn()
+        .expect("could not open to specified directory");
+
+    #[cfg(target_os = "macos")]
+    Command::new("open")
+        .arg(dir)
+        .spawn()
+        .expect("could not open to specified directory");
+
+    #[cfg(target_os = "linux")]
+    Command::new("xdg-open")
+        .arg(dir)
+        .spawn()
+        .expect("could not open to specified directory");
 }
 
 #[tauri::command]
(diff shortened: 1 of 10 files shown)


src/components/mainMenu.vue
```diff
@@ -113,7 +113,7 @@ export default defineComponent({
       try {
         const log_path = await util.getLogPath();
         util.infoLogger('log path acquired:' + log_path);
-        await shell.open(log_path);
+        tauri.invoke("open_folder", {dir: log_path});
```
isSerge, line 116: `invoke` is also async

src/lib/config.ts
```diff
@@ -115,13 +115,6 @@ class Config {
    * @param {IConfig} - config object to store as config file
    */
   private async write(config: IConfig): Promise<void> {
```
isSerge, line 117: where do we create a directory instead?

src/components/mainMenu.vue
ozgunozerk: nice catch, thanks

src/lib/config.ts
ozgunozerk: https://github.com/subspace/subspace-desktop/pull/342/files/4271bed595cb7af0703640e21551d7225adb2faf#diff-0b4f5977a52cef50ed20e7cdf9761795df9ee1cb3b5e5ea45edeb9394e8a47ffR65

here, trail of bits reported we are creating the config file in 2 different locations, so I kept the one in `init()`

src-tauri/src/utils.rs
```diff
@@ -9,6 +10,27 @@ pub(crate) struct DiskStats {
     total_bytes: u64,
 }
 
+#[tauri::command]
+pub(crate) fn open_folder(dir: String) {
+    #[cfg(target_os = "windows")]
+    Command::new("explorer")
+        .arg(dir)
+        .spawn()
+        .expect("could not open to specified directory");
```
achiurizo, line 19:
```suggestion
        .expect("could not open the specified directory");
```

_very_ minor nitpick feel free to ignore 😄

achiurizo (approved):

src-tauri/src/utils.rs
ozgunozerk: lemme fix that in the other pr, thanks for noticing that!

PR #348, opened 2022-09-27
Fix/update default plot directory

natachadelarosa:


ozgunozerk (changes requested):
1- With the new changes, the check on line 262 (https://github.com/subspace/subspace-desktop/pull/348/files#diff-e879b9701cf110ae7a5636e48ba2803b2f6a8b808b3539e2937607f7a0200d11R262) won't be necessary anymore. We can simplify that check

2- Here is the bug introduced with these changes:
- default plot path is not `previous path / plots`, which is a non-existent directory.
- this directory will be created when user confirms to proceed to plotting in the `SetupPlot` page.
- However, because this directory not exist whilst the user is on the `SetupPlot`, when user clicks the `folder` icon, and then `cancel` the dialog box back again, the backend function `get_disk_stats` from the `utils.rs` gets triggered. And crashes the application
- the reason is, `get_disk_stats` is taking a directory as its parameter, and when this directory is non-existent, it crashes, naturally.

We have to call `get_disk_stats` after the `cancel` action, since the user may have selected another disk. So it's not an option to: *not call `get_disk_stats` after `cancel`*.

ozgunozerk:
My suggestion would be (for the 2nd item):

immediately create the `plot` directory if it is not found. We are working in the `subspace-desktop` directory nevertheless, so creating a folder inside that folder, I don't think we need an explicit permission from the user for that. That wouldn't make our users angry imo :)

ozgunozerk:
Another suggestion would be, if plot directory does not exist yet, and we don't want to create it without prompting the user. Then, we can supply the `subspace-desktop` folder (1 level upper in this case compared to `plots`) to the `get_disk_stats` call instead. That would also solve the problem.

ozgunozerk:
Thanks @natachadelarosa ! 
I think this is good to go for the user's perspective. 

There is 2 small issues left imo:
1- when the application proceed into `SetupPlot` we are getting this error on the console
image

2- And when the folder icon is clicked, we are getting the last error on the console
image

@isSerge did a great job cleaning all the **red** entries on the console, keeping it clean as it is may be a good idea :)

src/pages/SetupPlot.vue
```diff
@@ -309,6 +301,15 @@ export default defineComponent({
       if (result) {
         this.store.setPlotPath(result);
       }
+
```
ozgunozerk, line 304: I see 2 problems with this approach. 
1- there are still `red` lines in the console, see my above comment (case 2: when folder icon is clicked although there is no `plot` file created)
2- this function now has the privilege to create folders, and this creation operation is applied after the user selects directory... If we are not in the `subspace-desktop` folder, I think we should not create any folders without prompting the user. 

My suggestion would be: moving this chunk of code into the `mounted()` part, or maybe creating a separate function for these chunk of code and calling that function in the `mounted()` part, so that we know we will only create the `plot` folder, nothing else. And other functions will stay clean

ozgunozerk (changes requested):

src/pages/SetupPlot.vue
```diff
@@ -230,6 +230,7 @@ export default defineComponent({
     await this.updateDriveStats();
     const path = (await tauri.path.dataDir()) + util.appName + util.PLOT_FOLDER;
     this.store.setPlotPath(path);
+    this.createDefaultPlotDir();
```
ozgunozerk, line 233: `await`? 
other than, this looks great to me, I don't have any more suggestions :)

ozgunozerk (changes requested):

ozgunozerk (approved):

PR #349, opened 2022-09-29
feat(config): create from backend

dnoishi:


src/lib/util/tauri.ts
```diff
@@ -54,3 +54,13 @@ export async function resetAndClear({
...
 }
+
+export async function writeFile(
+  configFullPath: string,
+  config: IConfig
+): Promise<void> {
+  await invoke('create_file', {
+    path: configFullPath,
+    content: JSON.stringify(config, null, 2)
+  }).catch((error) => errorLogger(error));
```
isSerge, line 65: I am changing error handling in another PR, but moving forward instead of catch and log it is better to throw an error and update store, so it is displayed on UI

ozgunozerk:
I'll look into failing tests now, seems like its about the new dependency on the rust side

src-tauri/src/utils.rs
```diff
@@ -77,3 +80,16 @@ pub(crate) fn custom_log_dir(id: &str) -> PathBuf {
...
+// TODO: current implementation works only on unix like systems
+#[tauri::command]
+pub(crate) fn create_file(path: &str, content: String) {
+    let mut file = File::create(path).expect("can't create file");
+    file.write_all(content.as_bytes())
+        .expect("couldn't write file");
+    let metadata = file.metadata().expect("can't get metadata");
+    let mut permissions = metadata.permissions();
+
+    permissions.set_mode(0o644); // Read/write for owner and read for others.
```
ozgunozerk, line 93: why `644`? Do we want it to be readable by the group and public?

ozgunozerk: I think it's better we disable read for others as well, no need for a 3rd party to reach the reward address (although it is not critically harmful, there is simply no need)

ozgunozerk (changes requested):

src-tauri/src/utils.rs
```diff
@@ -77,3 +80,16 @@ pub(crate) fn custom_log_dir(id: &str) -> PathBuf {
...
     path.expect("log path generation should succeed")
 }
+
+// TODO: current implementation works only on unix like systems
+#[tauri::command]
+pub(crate) fn create_file(path: &str, content: String) {
+    let mut file = File::create(path).expect("can't create file");
+    file.write_all(content.as_bytes())
+        .expect("couldn't write file");
+    let metadata = file.metadata().expect("can't get metadata");
```
ozgunozerk, line 90:
```suggestion
        file.metadata()
        .unwrap_or_else(|why| panic!("couldn't get metadata, because: {why}"))
        .permissions()
        .set_mode(0o644);
```

I think this would be more idiomatic, we don't need to store intermediate steps in variables. 
`assert` was a nice touch, thanks for that! We should probably use these in the tests when we have them.

src-tauri/src/utils.rs
ozgunozerk, line 93:
```suggestion
        file.metadata()
            .unwrap_or_else(|why| panic!("couldn't get metadata, because: {why}"))
            .permissions()
            .set_mode(0o644);
```

I think this would be more idiomatic. We don't need to store the intermediate computations in variables for now. 
`assert` was a nice touch! We probably should use it when we have tests for the backend. Thanks!

src/lib/util/tauri.ts
ozgunozerk: We can update here accordingly if your PR merges first

ozgunozerk:
@natachadelarosa 
backend is finished :)
after Serge's PR on errors, conflicts needs to be resolved, and the error handling for config on the frontend probably needs to be revisited. 
Do you want to handle those?

src/lib/util/tauri.ts
```diff
@@ -50,3 +50,13 @@ export async function resetAndClear({
   // this may throw error that file does not exist if config file is in the same directory as plot files (removed on the line above)
   await config.remove();
 }
+
+export async function writeFile(
```
isSerge, line 54: it seems we only use it in the `Config` class, so maybe we could use `invoke` inside `Config.write` instead of declaring this funciton in `utils` and injecting it into `Config`?

src/lib/config.ts
```diff
@@ -115,9 +115,8 @@ class Config {
    * @param {IConfig} - config object to store as config file
    */
   private async write(config: IConfig): Promise<void> {
-    return this.fs.writeFile({
-      path: this.configFullPath,
-      contents: JSON.stringify(config, null, 2)
+    return await this.writeFile(this.configFullPath, config).catch(error => {
```
isSerge, line 118: I think it can be just `return this.writeFile(this.configFullPath, config)` or `invoke(...)` directly without `.catch` and throw

src/lib/util/tauri.ts
isSerge: Nevermind, we can keep it separate and inject (to keep it testable)

src/lib/config.ts
ozgunozerk: I agree, that was also what I had in mind, and to my tests, it works like that

isSerge (approved):

ozgunozerk (approved):
LGTM

PR #351, opened 2022-09-30
validate farmer data at the beginning

ozgunozerk:
@nazar-pc , trail of bits suggested doing these checks at the beginning of the function to reduce side-effects. Thought it might be helpful to the monorepo as well


nazar-pc (approved):
Sure, why not

i1i1 (approved):

PR #352, opened 2022-09-30
Trail of bits audit - Issue #3 (Improper error handling)

### LINKED ISSUE #337: Trail of Bits audit - Issue #3 (Improper error handling)
The code does not change its behavior when errors occur in the following cases:

- [x]  The Linux auto launcher createAutostartDir function does not return an error if
it fails to create the auto start directory.
- [x]  The Linux auto launcher enable function does not return an error if it fails to create
the auto start file.
- [x]  Linux auto launcher disable does not return an error if it fails to remove the auto
start file.
- [x]  Linux auto launcher isEnabled will always return true even if it fails to read the
auto start file, which indicates that the auto launcher is disabled.
- [x]  The exportLogs function does not display an error message to the user when it
has an error. Instead, it silently fails.
- [x]  The startFarming checks if rewardAddress is set, sends an error log to the back
end, but still tries to start farming without a reward address, causing the back end
to error out without a clear message displayed in the front end.
- [x]  The Config::init function does not show the user an error message when it fails
to create the config directory.
- [x]  The Config::write function does not show the user an error message when it
fails to create the config directory and proceeds to try to write to the non-existent
config file. It also does not provide an error message when it fails to write the config
file in a call to writeFile.
- [x]  The removePlot function does not return an error if it fails to delete the plots
directory.
- [x]  The createPlotDir function does not return an error if it fails to create the plots
folder (e.g., if the user does not have permission to create the folder in that
directory). This will cause the startPlotting function to fail silently without the
user understanding what the problem is.

isSerge:
Introduced new error titles:
```
"Failed to initialize autolauncher",
"Failed to initialize updater",
"Failed enabling autolauncher",
"Failed disabling autolauncher",
"Failed to get logs",
"Failed to initialize config",
"Failed to reset the application",
"Failed to start plotting"
```

Draft, since we need to update the current generic error message:
```
Please, make sure you have an Internet connection and enough disk space in order to proceed. If the problem persists, you can reach out to our team
```

Closes #337


src/components/mainMenu.vue
```diff
@@ -83,9 +83,17 @@ export default defineComponent({
         });
       }
       if (this.launchOnStart) {
-        await this.$autoLauncher.enable();
+        try {
+          await this.$autoLauncher.enable();
+        } catch (error) {
+          this.store.setError({ title: 'errorPage.enableAutoLauncherFailed' });
```
ozgunozerk, line 89: do we need this? I assumed the error title set and error handling was done in the  `src/boot/init.ts`. Won't this be a duplicate code?

src/components/mainMenu.vue
```diff
@@ -83,9 +83,17 @@ export default defineComponent({
...
+        }
       } else {
-        await this.$autoLauncher.disable();
+        try {
+          await this.$autoLauncher.disable();
+        } catch (error) {
+          this.store.setError({ title: 'errorPage.disableAutoLauncherFailed' });
```
ozgunozerk, line 95: same for here

src/i18n/en.json
```diff
@@ -136,6 +136,14 @@
     "startNodeMissingParams": "Node name and plot directory are required to start node",
     "startFarmerFailed": "Failed to start farmer",
     "defaultErrorMessage": "Please, make sure you have an Internet connection and enough disk space in order to proceed. If the problem persists, you can reach out to our team",
-    "defaultErrorTitle": "Something went wrong"
+    "defaultErrorTitle": "Something went wrong",
+    "initAutoLauncherFailed": "Failed to initialize autolauncher",
+    "initUpdaterFailed": "Failed to initialize updater",
+    "enableAutoLauncherFailed": "Failed enabling autolauncher",
+    "disableAutoLauncherFailed": "Failed disabling autolauncher",
```
ozgunozerk, line 143: nitpick, feel free to ignore:
for consistency I suggest `Failed to enable autolauncher`, and `Failed to disable autolauncher`

src/lib/autoLauncher.ts
```diff
@@ -43,50 +43,40 @@ export class LinuxAutoLauncher {
...
     response.stdout.push('success');
     return response;
   }
 
   public async disable(): Promise<ChildReturnData> {
-    const autostartAppFile = await this.getAutostartFilePath();
+    const autostartAppFile = this.getAutostartFilePath();
     const response: ChildReturnData = { stderr: [], stdout: [] };
-    await this.fs.removeFile(autostartAppFile).catch((error) => {
-      errorLogger(error);
```
ozgunozerk, line 57: why are we removing `catch(error)`? Are we sure it won't produce any errors? I see in many of the below changes, `catch(error)` is deleted along with the `try` block. Couldn't understand what's happening tbh

ozgunozerk (commented):
I noticed this PR will get rid of `errorLogger`, but we also wan't errors to be logged. 
When I looked at `setError()` code in `store.ts`, I couldn't see any call to `errorLogger()`?

src/components/mainMenu.vue
isSerge: init is separate

src/lib/autoLauncher.ts
isSerge: in order to propagate error, otherwise, `catch` will swallow it

ozgunozerk:
@isSerge I'm happy with it as is. However, what about embedding `errorLogger(error)` into the `setError(error)` function, so that we will only call `setError()` from the outside?
The code may be cleaner this way.

isSerge:
@ozgunozerk `setError` does not have access to the logger, unfortunately, it seems Pinia does not support dependency injection, so we'll have to provide it as a parameter as in other methods

isSerge:
@ozgunozerk also `setError` accepts objects with a title and optional message (which are displayed to users), while the logger logs the actual error, so I would prefer to keep them separate

PR #357, opened 2022-10-06
Frontend / backend privilege separation

### LINKED ISSUE #339: Trail of Bits audit - Issue #5 (Insufficient privilege separation between the frontend and backend)

ozgunozerk:
I've replaced the frontend `fs` commands with the backend ones (implemented backend ones for this). 
@isSerge fixed the unit tests and went over the frontend.

### This is a big PR, here is the detailed summary:

for each generic filesystem related function (read, write, create, remove), I've found the specific use case and tried to narrow down the scope. 

changes:
- `readConfigFile` -> reading the config file is now handled by the backend. This is now a `tauri` function.

- `getLogPath` and `invoke(open_folder)` are merged into -> `openLogDir()`
    - now, we don't have a generic `open_folder(path)` function, which allows to open any path in the filesystem
    - `openLogDir()` opens the log location only, does not take a `path` parameter
    
- `writeFile(path, contents)` is abandoned
    - **use case 1**: creating the autolauncher for linux:
        - we now have `createLinuxAutoLaunchFile` invokable from tauri, which does not take any parameter for the path
    - **use case 2**:  write/create the config file
        - we now have `writeConfig(contents) invokable from tauri, which does not take any parameter for the path
        - this function needs to take `rewardAddress` and other fields as parameter, so for an XSS attack, changing the reward address using this application is still possible. But I guess there is nothing we can do against it. Frontend should have access to manipulate the reward address.
        
- `removeFile(path)` is abandoned
    - **use case 1**: removing the autolauncher for linux:
        - we now have `removeLinuxAutoLaunchFile` invokable from tauri, which does not take any parameter for the path
    - **use case 2**: removing the config
        - we now have `removeConfig()` invokable from tauri, which does not take any parameter for the path
        
- `isDirExists(path)` -> this is generic, but only returns `true/false` depending on the given folder exists. I believe this is safe.

- `readFile(path)` is abandoned
    - **use case 1**: reading the linux auto launch file to see whether it exists
        - we now have `linuxAutoLaunchFileExist` invokable from tauri, which does not take any parameter for the path
    - **use case 2**: reading the config to retrieve its contents
        - we now have `readConfig()` invokable from tauri, which does not take any parameter for the path
        
- `createDir(path)` -> this is also generic, and creates an empty directory at the given path. I believe it is not a security concern to be able to create empty directories.

- `removeDir(path)` -> this one is tricky. `reset` button needs to clear all plot files. And plot files can be placed in a custom folder by the user. So, this function needs the `path`. Which means, backend is able to delete ANY folder recursively. To prevent this, I've found the following solution:
    - the folder to be recursively deleted is supposed to be a folder that contains data ONLY related to subspace.
    - the folder (which is to be recursively deleted), is now getting inspected by the backend, and if path does not end with `plots`, the operation is aborted.

P.S. all the invoke functions are documented

Fixes #339


isSerge:
I am still considering some improvements on the frontend, converting to draft for now

ozgunozerk:
After testing all these, let's make this pr ready :)

ozgunozerk:
Tested the config and autostart related functionalities on:
- [ ] macos
- [ ] windows
- [ ] linux

It's good to go imo

isSerge (approved):

